PR #39, opened 2017-04-04
Take over some of saibotu's improvements

d0k3:


source/fs/fsutil.c
```diff
@@ -525,11 +525,9 @@ bool PathCopyVrtToFat(char* dest, char* orig, u32* flags) {
...
     }
     
-    // check destination write permission (special paths only)
-    if (((*dest == '1') || (strncmp(dest, "0:/Nintendo 3DS", 16) == 0)) &&
-        (!flags || (*flags & (OVERWRITE_CUR|OVERWRITE_ALL))) &&
-        (!flags || !(*flags & OVERRIDE_PERM)) && 
-        !CheckWritePermissions(dest)) return false;
+    // check destination write permission
+    if ((!flags || !(*flags & OVERRIDE_PERM)) && !CheckWritePermissions(dest))
+        return false;
```
d0k3: A quick note, just in case you wonder. I took over the commit before this to my GM9 repo, and part of this commit.

The reason why I didn't take over this (f.e.) is because what you removed here is actually a somewhat hacky solution to speed things up, especially when copying large numbers of files. The addditional checks in here make sure that we only check write permissions when it actually makes sense (on drive 1:/ and in the 0:/Nintendo 3DS folder).

If you ever intend to do a pull request, unsure about the features that come in the commits after this. Can be discussed, though. If this was mainly for your personal use, don't bother ;).

saibotu: I only intended to do pull requests for this commit and the previous one. That's why i had them in two seperate "pr-" branches.

I moved the permission check from PathCopy into the worker, because i failed to realize it will be called for every single file when copying a folder. When copying single files (which is the only thing it tested) this saved a permission check in most cases.

d0k3: Yup, I noticed. To be completely honest, I don't even know if checking for each and every file would cost that much time. Better not to do it, though, if we can prevent it. In any case, thank you for these improvements! If you want your name on that second commit, I'd need your Github email so that I can attribute it back to you. You can also send this in a pm, though.

PR #41, opened 2017-04-08
Fixed ticket.db parsing

saibotu:
This fixes the offsets of tickets in the ticket.db virtual drive. All tickets had the same content without this.

I also added .tik extensions, so the tickets can be installed using FBI.

### DIFF
--- a/source/virtual/virtual.h
+++ b/source/virtual/virtual.h
@@ -28,7 +28,7 @@
 // bits 16...21: reserved for external flags
 // bits 22...31: reserved for internal flags (different per source)
 typedef struct {
-    char name[32];
+    char name[48];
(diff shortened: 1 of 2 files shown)


d0k3:
Goddammit, thanks for this pull request. Coincidentially you help me fix another bug I've been puzzled about for the last few hours with this. @saibotu - you're the best!

However, that other commit - problem is, this increases the size of of the VirtualFile struct, which might lead to unexpected problems elsewhere. I'll have to look into this further. I think there could be problems in vgame.c

d0k3:
Okay, as you may have noticed, I changed the second commit just a little. I think it is for the better this way, just to be safe. Again, thanks for your help!

PR #45, opened 2017-04-13
increase ramdrive storage size, reorganized the memory layout

Wolfvak:
Hasn't been heavily tested (I don't even have a N3DS to test the upper 128MiB) but it appears to work fine.


Ordim3n:
Works perfectly with N3ds, filled the ramdrive to full capacity and SHA matched. And while the ramdrive was full, i made a cia out of a random system title and mounted ticket.db as a virtual drive

PR #56, opened 2017-04-30
Update README.md to Luma3DS 7.0

Multipixelone:


d0k3:
I'd rather keep this in a state where instructions are correct for newer *and* older versions of Luma 3DS.

PR #73, opened 2017-05-26
Remove firmtool submodule

TuxSH:
This remove the `firmtool` submodule.

`firmtool` is an installable script and should be treated as such.

People who can run the script are also able to do `pip install --upgrade git+https://github.com/TuxSH/firmtool.git` to install (or update) `firmtool` easily.

Morever this avoids you from having to update the submodule everytime, and more importantly from keeping an outdated (and/or bugged) version of it in your repo. The submodule commit was older that v1.0 (for example it generated .sha files everytime, something you probably don't want)!


fox8091:
Perhaps also fix the build process so it uses an ELF instead of a bin since firmtool supports that?

d0k3:
Taken over locally

PR #83, opened 2017-06-04
Fix TWL key 0x03 init

TuxSH:
This fixes TWL key init:

* only done on k9lh
* fixed for devunits


d0k3:
Taken over locally, key0x03Y pulled out of P9 instead of hardcoded. You may review the changes here:
https://github.com/d0k3/GodMode9/commit/a27bd1f82d4da1cab673d0622eca9fb3dfe0f0d6

Also, I trust this is actually tested on A9LH?

TuxSH:
`Also, I trust this is actually tested on A9LH?`

I haven't tested on A9LH, although it _should_ work (the full routine which doesn't assume a9lh works).

PR #86, opened 2017-06-04
Drop support for pre-a9lh entrypoints

TuxSH:
Those entrypoints are terribly outdated and shouldn't be used anymore outside sb9si.

As you can see this also removes the need for painful-to-maintain support cruft inside GM9.


source/fs/fsperm.c
```diff
@@ -171,7 +168,7 @@ bool SetWritePermissions(u32 perm, bool add_perm) {
                 return false;
             break;
         case PERM_SYS_LVL3:
-            if (!ShowUnlockSequence(6, "!THIS IS YOUR ONLY WARNING!\n \nYou want to enable SysNAND\nlvl3 writing permissions.\n \nThis enables you to OVERWRITE\n%s", IS_SIGHAX ? "your B9S installation and/or\nBRICK your console!" : IS_A9LH ? "your A9LH installation and/or\nBRICK your console!" : "essential system files and/or\nBRICK your console!"))
+            if (!ShowUnlockSequence(6, "!THIS IS YOUR ONLY WARNING!\n \nYou want to enable SysNAND\nlvl3 writing permissions.\n \nThis enables you to OVERWRITE\n%s", IS_SIGHAX ? "your B9S installation and/or\nBRICK your console!" : "your A9LH installation and/or\nBRICK your console!"))
```
dark-samus: I think there's a typo here

AnalogMan151: Is the typo that this B9S only version is still referencing A9LH even though if they had it this version could not be run?

mariogamer2: It's still have a9lh binary @AnalogMan151

AnalogMan151: My apologies, then I cannot spot the typo.

mariogamer2: Me too...

d0k3:
While this is appreciated, it would be much more of a hassle to bring back removed entrypoints than  to lug them around for a little more.

Thus, this instead of taking the pull request:
https://github.com/d0k3/GodMode9/commit/d97e754593b31ed1618272aca67dca883c43890f
This will be kept until the next release to see if there are any legit complaints.

Please leave this pull request open, I'll take it at a later point.

linouxis9:
I wonder why they want payloads for old entrypoints be removed that badly..

d0k3:
There are good reasons to get rid of these unstable entrypoints. Difficult to maintain and test... Outdated, a lof of functionality not being available. There may also be legit reasons to keep, so I'm doing this slow.

fox8091:
Perhaps also remove usage instructions for old entrypoints from the readme as well?

d0k3:
Entrypoints removed locally, thus closing this now. Thanks for this pull request!

PR #97, opened 2017-06-14
Wait 3ms for i2c reads/writes to workaround N3DS MCU bugs

Hikari-chin:
Luma, b9s, etc. wait 3ms when doing i2c reads/writes due to MCU bugs on the N3DS, however GM9 does not. This fixes that.

### DIFF
--- a/source/common/i2c.c
+++ b/source/common/i2c.c
@@ -1,5 +1,7 @@
 #include "i2c.h"
 
+void wait(u64 amount);
+
 //-----------------------------------------------------------------------------
 
 static const struct { u8 bus_id, reg_addr; } dev_data[] = {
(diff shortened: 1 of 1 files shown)


d0k3:
Okay, I have a suggestion for this one. Maybe @AuroraWright can chime in, too.

These delays, on all reads and writes, can lead to less responsive input routines (even if it is just 3ms). As I understand, MCU bugs on N3DS only happen when alternating between reads / writes, though. So... what if we add the 3ms delay *only at the beginning and end of i2cWriteRegister()*. That should work, too, right?

d0k3:
Merged locally into the scripting branch, which will be merged to master later. Thank you!

fox8091:
This uhh..didn't actually do anything. Luma defines wait() in utils.c as
`
void wait(u64 amount)
{
    startChrono();

    u64 initialValue = chrono();

    while(chrono() - initialValue < amount);
}`
Meanwhile, this is simply a blank function.

Hikari-chin:
Forgive me. It slipped my mind that utils.c was a Luma-only thing. I'll fix it tomorrow.

PR #122, opened 2017-07-17
devkitARM r47 fix

WinterMute:
inline alone leaves the compiler free to use the external definition, but the program does not provide one.


d0k3:
Merged

PR #133, opened 2017-07-29
Update hid.h

emre009:
https://github.com/emre009/GodMode9/commit/f3f79dac9556e23ec9f5a71229da9dcbac0e8ce4

### DIFF
--- a/source/common/hid.h
+++ b/source/common/hid.h
@@ -26,7 +26,7 @@
 
 // special buttons / cart / sd
 #define BUTTON_POWER  (1 << 12)
-#define BUTTON_HOME   (1 << 13)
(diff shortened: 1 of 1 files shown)


ahoischen:
1. This change caused GodMode9 to break.
2. As repeatedly stated, special "broken button" builds aren't being made. You can change the code yourself, but don't expect it to be incorporated into the central codebase.

mariogamer2:
No you must also switch what is in parentheses. Also @ahoischen second point

d0k3:
What @ahoischen said. This special broken button build is for you, it won't be incorporated into the codebase. Think about it, @emre009 - the next guy has a broken Y button, and what then?

PR #144, opened 2017-08-07
Fixed mounting images

### LINKED ISSUE #142: GM9 1.3.0 brakes FAT Image Mount at DSiware sav file
Yep, pretty much.
UPDATE: Old GM9 works fine

saibotu:
This fixes mounting images (closes #142), while still allowing mounting from 8: and 9: (except for nand images).

There might be a better solution to this, but i couldn't find a way to identify the amount of needed drives and didn't know where to implement it.

### DIFF
--- a/source/filesys/fsinit.c
+++ b/source/filesys/fsinit.c
@@ -35,13 +35,12 @@ bool InitExtFS() {
 
 bool InitImgFS(const char* path) {
     // find drive # of the last image FAT drive
+    u32 type = IdentifyFileType(path);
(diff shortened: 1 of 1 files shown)


saibotu:
Don't merge this yet, just noticed another issue with this. Mounting a nand image and then any other image doesnt restore ramdrive and bonusdrive.

d0k3:
Yeah, that's correct. I didn't see your p/r earlier, so I just did the solution myself. Maybe you doublecheck?

saibotu:
Just checked, your solution works fine. 👍

PR #149, opened 2017-08-08
Added brightness adjustment through the volume slider

Wolfvak:
Also adds some ARM-specific routines, GIC/DIC code for the MPcore and PXI support.
The startup brightness has been lowered considerably (75% -> 25%) to prevent headaches and such.


d0k3:
Taken over locally, thanks!

PR #153, opened 2017-08-12
Enable alternate key .firm build (for ntrboothax, etc)

rboninsegna:
:3


Wolfvak:
Could you be more specific as to which `minor features` are unavailable when booting from this entrypoint?

As for the lack of decrypted OTP, is it because the necessary key is cleared before boot or is there any other reason?

d0k3:
No OTP access, @Wolfvak :)

Wolfvak:
Merged with https://github.com/d0k3/GodMode9/commit/e95ece09f4d85260d522ba95d10a9af7517aedd4, thanks for the contribution!

PR #154, opened 2017-08-12
Pass argc=1 if the specified FIRM does not require screen init

AuroraWright:
For consistency with B9S/Luma3DS


d0k3:
Taken over locally, thanks!

PR #163, opened 2017-08-15
Add local exception handler, improve MPcore IRQ handler

Wolfvak:
The ARM9 exception handler only modifies a small section at the bottom of ITCM, so any other global handlers that may be present won't be affected.

The ARM11 IRQ handler was changed to use some new ARMv6k instructions.


d0k3:
Taken over locally, thanks!

PR #168, opened 2017-08-16
Add Nintendo Zone to injection list

shoguna:
Since the app is completely inoperable even when you manage to launch it from FBI, could you add Nintendo Zone to the list of Nintendo apps to inject?


ahoischen:
This seems to be an issue and not a PR. Correct me if I'm wrong, but the proposed changes don't seem to be related to your request. Consider opening an issue instead.

PR #191, opened 2017-09-02
Scripting: SD card un/remount

### LINKED ISSUE #190: Feature Request - Script support for mounting/unmounting SD
With NTRBoothax's release, many people are trying to make installation packs for use with NTRBoothax. These packs usually include some pre-downloaded files like boot9strap, Luma3DS, FBI, etc and the whole process is executed in a set of scripts. One key part is the copying of setup files from the installation SD card to the RAMDISK, then back again to the target SD card after unmounting and swapping the SD cards. Since unmounting cannot currently be done by script, this means either breaking it up into two scripts or removing the SD card unsafely. 

Since scripts run in memory, I would like to ask if scripting commands for `InitSDCardFS()` and `DeinitSDCardFS()` could be added so we could move files between SD cards using the RAMDISK without the need of multiple scripts.

zetaPRIME:
Allows for SD card swapping during script execution. Would resolve #190.

### DIFF
--- a/source/filesys/fsscript.c
+++ b/source/filesys/fsscript.c
@@ -41,6 +41,8 @@ typedef enum {
     CMD_ID_MKDIR,
     CMD_ID_MOUNT,
     CMD_ID_UMOUNT,
+    CMD_ID_SDMOUNT,
+    CMD_ID_SDUMOUNT,
     CMD_ID_FIND,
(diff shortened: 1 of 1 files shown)


d0k3:
Seen this too late, already solved it. You forgot setting the EmuNAND offset, though :).

PR #193, opened 2017-09-05
Fix failure deleting/renaming mounted files

windows-server-2003:
It will show no confirmation to unmount the file and fail when try to delete/rename the mounted file via X or R+X,so I added confirmation to unmount the mounted file when deleting/renaming.


windows-server-2003:
Sorry,it isn't so easy.I didn't test enough.Folder deletion will be bugged.I will close this.

d0k3:
Exactly, that's the problem. And there is no easy / quick way of handling this yet.

PR #206, opened 2017-09-18
Typo fix

DMSalesman:
Fixed a minor typo.


d0k3:
I merged it together with another small readme fix of mine (so not to clutter up commit history) - hope that's okay for you!

DMSalesman:
You are more experienced with GitHub than me, plus it's your project we're talking about, so the rules are yours to decide! I only care that the result be as perfect as possible.

PR #231, opened 2017-10-04
SALTMODE

RShadowhand:
This mode changes GM9 bootloader mode to behave more like an actual bootloader.
Changes:
- No splash.
- No 500ms delay before boot for splash.
- Loading process won't stop because you held a key (other than START), making it possible to get into Luma's config menu in the off chance you need to change something).
- Hold START to get to the menu, instead of R+LEFT.

Compile using `make firm SALTMODE=1`


d0k3:
Taken over locally (with some small adaptions)

PR #232, opened 2017-10-07
add 'unzip' command

windows-server-2003:
I added "unzip" command for extracting noncompressed zip file.
Please correct my bad English in source.
And also I couldn't add zip extraction via file handler menu. so if you can, please add. (sorry...)


windows-server-2003:
TravisCI building failed, but don't mind. The problem was the server's downloading and built successed in local.

windows-server-2003:
Close with this conversation
[https://gbatemp.net/threads/release-godmode9-all-access-file-browser-for-the-3ds.420027/page-33](url)

PR #249, opened 2017-10-25
Added REGION variable for scripting (fix #243)

### LINKED ISSUE #243: Add $[REGION] var to GM9 script engine
I already brought this up on IRC and you agreed it'd be a good idea, so I figured I'd drop it here until it's implemented. 

For everyone else:
Simple, do a check to see what region the 3DS is or what region the NAND currently is at least (in case the user region changed?) and set the var to one of the 6 major regions based on that.
JPN, EUR, USA, CHN, KOR, TWN

The use case I listed is for filenames, say I wanted to make a luma 3DS autosetup script, I could include 6 config files, one for each region named as config.USA.bin, config.EUR.bin, config.JPN.bin, etc. and in the script, I could use 
```cp -o -n G:/config.$[REGION].bin 1:/rw/luma/config.bin```
to grab the correct file based on the region of the system and copy it.

There's probably other use cases too, and this will also be useful if #129 is ever implemented, since you could perform operations based on what the region of the console is.

figgyc:
I tested it and it worked. Fairly simple change.

### DIFF
--- a/HelloScript.gm9
+++ b/HelloScript.gm9
@@ -29,6 +29,7 @@ ask "Continue running this script?"
 ask -o -s "Really continue running this script?\n(I will completely ignore your answer)"
 
 # ENVIRONMENTAL VARS
+# REGION is the region of your device's SysNAND (USA, EUR, JPN, KOR, CHN, TWN or UNK for unknown)
 # SERIAL is the serial number of your device
 # GM9OUT is the standard output path
 # CURRDIR is the directory the script is running from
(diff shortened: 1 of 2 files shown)


d0k3:
While your pull request is good code, I had to change some stuff, to prevent duplicate code and bloat. The SecureInfo file is already loaded for the SERIAL env var, a little above, and it's better region is handled in the same place.

Anyways, taken over locally, thank you!

PR #257, opened 2017-11-04
Add way to change file attributes

angelsl:
Took me a while. Sorry about that.

### DIFF
--- a/source/filesys/fsutil.c
+++ b/source/filesys/fsutil.c
@@ -666,6 +666,11 @@ bool PathRename(const char* path, const char* newname) {
     return (f_rename(path, npath) == FR_OK);
 }
 
+bool PathAttr(const char* path, u8 attr, u8 mask) {
+    if (!CheckDirWritePermissions(path)) return false;
+    return (f_chmod(path, attr, mask) == FR_OK);
+}
+
 bool FileSelector(char* result, const char* text, const char* path, const char* pattern, bool hide_ext, bool no_dirs) {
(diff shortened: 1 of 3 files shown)


angelsl:
Hold on, don't merge yet.

Need to not show the attribute functionality for virtual files.

Also, thoughts on the interface itself?

angelsl:
Okay. It no longer shows options to change on virtual files (and directories).

It works on directories now too.

d0k3:
Alright, interface looks okay. You seem to have broken something, though. Check long file / dirnames.

The other pull request I'll take.

angelsl:
Alright, updated.

The menu is only for files now. Attributes are set on exit, and the user is asked to confirm the changes.

d0k3:
Okay, I'm nitpicking here, but, two things...

1. (press arrow keys to change attributes) -> make this short, this line shouldn't be longer than the above block ([arrow symbols] to change attribute) would be enough (whereas [arrow symbols] are the actual arrow symbols).
2. The confirmation dialogue will look bad for actual long filenames - gotta change that. Tbh, to me, that dialogue seems like overkill anyways. It would be enough if the bottom line said (<A> to apply, <B> to cancel), maybe even only after a change is detected.

Anyways, look into that, and I'll take it over.

d0k3:
Alright, I took it over locally, and added the two mentioned changes myself. I hope that's okay.

Normally I wouldn't push stuff that quick, but there are some bigger file reorganizations underway, so I wanted this pull request through before that happens.

PR #258, opened 2017-11-04
Timeout SD init after about a second

angelsl:
This fixes the hang on boot if there is an empty microSD to SD adapter inserted, or otherwise faulty SD card that won't complete initialisation properly.

(Now GM9 won't hang, but the FIRM after GM9 will hang... *sigh* why isn't there a common library for this ARM9 stuff yet)

### DIFF
--- a/source/nand/sdmmc.c
+++ b/source/nand/sdmmc.c
@@ -397,6 +397,7 @@ static int SD_Init()
     u32 temp = (handleSD.error & 0x1) << 0x1E;
 
     u32 temp2 = 0;
+    s32 timeout = 0x80;
     do
     {
         do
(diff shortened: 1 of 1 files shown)


angelsl:
Hmm, I think there's a better alternative to this.. not sure.

urherenow:
With an empty adapter inserted, will it continue to boot from firm if something is installed there, or will it still hang?

urherenow:
For example, you installed luma to firm so it boots without an SD.

angelsl:
> With an empty adapter inserted, will it continue to boot from firm if something is installed there, or will it still hang?

It will boot from FIRM0.

But if FIRM0 is something else that tries to mount SD and it doesn't have a time out, then that will hang.

angelsl:
> Hmm, I think there's a better alternative to this.. not sure.

OK, I [tried this](https://github.com/angelsl/GodMode9/commit/83052c3856ef6cbcc0027d94674867d87b4d4ebd) and it doesn't work too well (GM9 calls things that try to init SD three times before it hits the bootloader menu, so we end up having to abort thrice).

The timeout here needs to be adjusted too, right now it's a bit long because the SD takes a bit longer to init from cold boot.

d0k3:
Taken over. Given that this solves a very specific problem (one that *almost* falls under the "vandalism" category), and has no effect otherwise this is good enough. Thank you!

fox8091:
@d0k3 To be fair, I've used this before on an O3DS to allow me to leave the adapter in and just remove the microSD as needed. (Less wear on the 3DS SD reader, more on the easily replaceable adapter.)

d0k3:
Yup, that's okay. :)

Issue #129, opened 2017-07-23
Feature Request - Control Flow in GM9 Scripts


ValliantStorme:
Control flow commands would be nice (if-else, loops, and maybe jumps to marked locations in a script). 
I've been playing with the scripting feature all night, and so far I'm impressed with the amount of control this early iteration gives the user. I hope to see it expanded upon massively!

EDIT: Additionally, allowing a script to call another script would be nice too, but would probably be overkill or unnecessary.

EDIT 2: Didn't know GM9 supported file structures within /gm9/scripts. Good to know that it does, solved my problem with selection menus (for my use-case at least).

d0k3:
I'm actually thinking about adding a `goto` command, however this is prone to error and not a priority right now. Btw, I'm glad to see people putting scripting to use, and I hope to see some interesting scripts pop up soon :).

d0k3:
Arithmetic may be a good addition to this, too.

annson24:
Control Flow feature will be great, user input will be a good idea too rather than only having to choose from Yes/No(sorta).

Anyway, I'm up for testing.

jason0597:
Maybe just implement lua?

d0k3:
I guess you didn't think this day would come... thanks to the tireless work of @windows-server-2003 and in no small part thanks to [Kazuma77](http://gbatemp.net/threads/release-godmode9-all-access-file-browser-for-the-3ds.420027/page-43#post-7743068)'s testing, we now have `if`-`else`-`goto` in GodMode9 scripts. As a cherry on the Cake, I also added an `elif` command.

Thanks everyone involved, closing this now!

PR #261, opened 2017-11-07
add "if", "else", "end", "goto", "label" command

windows-server-2003:
The second commit includes the changes since you had a quick look.
solves #129

Edit
if I'm using bad English in the comments or error messages, please correct them.
I didn't add description in HelloScript.gm9 because if I write, it will not understandable.


d0k3:
Okay, doesn't look bad, I'm still hesitant to take this - this is critical stuff, so I need to fully understand how this works first.

Can you maybe give me a quick walkthrough of how this works? Also maybe some scripting examples?

d0k3:
Okay, after a slightly more thorough look...

static char lbl_table [_ARG_MAX_LEN + 1] [_MAX_LABELS];  // label table to save labels hisotry
I'd rather not do stuff like that ^ that's a shitton of memory you're trying to get there. It may not even be required to save this anywhere... you can just search for a label on demand and error out if not found. With the whole script being in memory and not above 1MB, that won't take too long. It wouldalso make the number of addditions to the source code smaller

Store pointers into the script instead - the label strings are already there, no need to store them in another place. In general, try to use as little memory as possible.

Btw, when I planned labels, I thought of something like this:
@label

The `label` command works, too, of course.

d0k3:
... aaand, just so you got notice in advance. There are some bigger file reorganizations underway, and there may be a merge conflict for your pull request afterwards.

That's nothing to worry about actually, if that happens you can just copy your scripting.c to the new location and reapply the commit to the new base.

windows-server-2003:
Wait a while. I'm trying to change label format to "@label".

d0k3:
Already looks a lot better!

Two questions: Why are static vars reset on line 1 (why not just before running the script?) and how is goto within if handled (if nesting).

windows-server-2003:
1)
Because their scope is only the function where they called(run_line())

2)
like this ?

if test test
    echo "0"
else
    label label
    echo "1"
end
goto label

This script will echo "0" then "1" then "1" ...(infinite loop)
Force don't skip after jump into "if" and after jump, everything will work normally

I found another issue.The script like below should exit with syntax error(duplicate "else"). but won't.
I can fix this, but for some reason, I can't use PC and git now and I don't know when restored.Sorry. I'll try to use git on Android.(yes, rooted)

goto label1
if test test
else
label label2
else
end
label label1
goto label2

windows-server-2003:
Fixed. but as I said, I can't build by myself now so if there are any bugs, sorry.
Used git on android(no root needed)

Quick description about the code attatched
[Quick.txt](https://github.com/d0k3/GodMode9/files/1456692/Quick.txt)

d0k3:
In any case, we'll wait until you can properly test. No merging before that. In the meantime, others can test, too (thanks to Kazuma77 of gbatemp fame here), and I can review the code.

windows-server-2003:
OK, I'll check a lot.

d0k3:
Alright, just a quick note, regarding radio silence. Once you have looked it over, get back, and I'll have another thorough look through the changes.

windows-server-2003:
Fixed the merge conflict.
I updated my branch "master" first, then fetched and merged "origin/master" into "if-goto"

d0k3:
Okay, after a thorough review. Being honest here, I still don't understand the code in full.

First of all - that code is still a bit hacky, and that in turn will make later additions (f.e. like a `for X in 0:/*.bin`) a lot harder than they ought to be (not that it would be too easy one way or another). `run_line` was only ever intended to process one unit of script code, with your additions the function is allowed to set the script pointer and besides that, does a lot more than just process one unit. I see that there is some trouble in processing if and the condition behind, of course. So, just a few ideas. Some of them might not be good, but everything can be discussed.

The search for a label should be done in a separate function, controlled by a global variable (the variable controlling it is already global, cause static == global). In `run_cmd`, everything that should be done for `goto` is set that global variable, everything that should be done for `label` is skip and return. If that global var is set, another function should reposition the pointer. In the same vein - you were right. A `label` keyword fits the current structure much better than the @ style labels I suggested.

Now, as for `if`, `else` and `end` control flow - I like what you did there with the if nesting. I do stumble over this, though: `{ CMD_ID_IF      , "if"      , 2, 0 },`. Why is the arg count here 2? If it's because most cmds take one argument: not all do. A solution for this would be putting the `if` condition into brackets and make the line parser recognize this as one argument. Might clash with some stuff, though, and we want to keep the error potential low. So, better not.

A different approach to this: make `if` have one argument. An `if` at the beginning of the line then has to be recognized specifically by `run_line` (that's the only one that absolutely has to). If `if` is recognized, you first make `run_line` execute what's behind the `if` silently and recursively (by calling itself with a change `line_start`. Then, depending on the result, you make `run_cmd` run with one argument, which is either `true` or `false` (or anything you choose, like `0` and `1`). This way, you can process stuff in `run_cmd`.

As for the search for next `else` / `end` on an `if false` - that can be handled similarily to the search for a label, outside of `run_line`. Same for the search for `end` after an else in case of an `if true`. You can, in fact, opt to ignore errors in parts of the script that don't get executed (you already do that in part).

tl;dr: use global vars and control the flow in external functions. Don't overload `run_line`. And get back about stuff you disagree with / and/or need clarification about.

windows-server-2003:
I understand.
Why is having two args bad? I think there are no reason to make "if" has one arg.
Also, I think "@" is better than "label". because special symbols used as special means(labels) makes scripts easy to read.

>> still a bit hacky, searching for the next else/end...
so I should make a var "skip_state" (0 if not skipping, 1 if "if" false and skipping until "else" or "end" , 2 if "if" true) and skipping until "end" and control it in run_cmd() ?

>> control flow in external function
Should I make a function similer to "run_cmd()"(more simple) for searching a label and leave work like calling parse_line() to run_line()? or do all stuff(includes calling parse_line()) in the external function?

>> Edit
how can I modify commits on the branch which opening PR? If I try to push rebased branch in local to the remote one, I get an error like "push rejected(your local branch is behind of the remote one)" ,so I had to delete my branch once, then push. But deleting the branch will close the PR(I heared), anyway I think there are some better ways to do that.

d0k3:
I modified your comment above a little, just so it is better readable.

Well, why does `if` need two args? The way I understand it (being honest, I don't fully understand your code here), it's if [argument], where [argument] is either true or false. [argument] has to be specifically handled by `run_line()`, of course, possibly via recursion as I suggested above.

I trust your judgement on most of that stuff, I just dislike these hacks in `run_line()`. `run_line()` has to be altered to handle `if`, but that's about it. It also should not be allowed to reposition the script pointer. `run_cmd` could handle these control flow commands via setting the global vars that control the flow (like, storing the label to search for). Neither `run_cmd` nor `run_line` should be allowed to reposition the script pointer.

I think you don't even have to change that much:
* `goto`can be handled as a standard cmd, `run_cmd` sets the global label var, which is empty unless we're searching for a label.
* The `if` condition can be handled recursively as described above.
* In `run_cmd`, `if` and `else` could alter some global vars that would tell another function were to go next or what to search for. Of course, it also makes sense to handle the if nesting in `run_cmd`.

And yes, I suggest you make a separate function that searches for the label on `goto`. Possibly also a separate one that handles `if` based script pointer changes (search for `else` / `end` on an "if false", f.e.).

The advantage of doing stuff this way is that the code in general will be easier to understand. `run_line` would not do more than you'd expect from the name (who would expect that to reposition the pointer?), and you'd have smaller, easier understandable units of code in general. 


As for your local branch - git is flaky sometimes. If the remote branch is up to date with the local one, just delete the local one and reclone. If the local one is farther ahead, store your changes somewhere locally, reclone, and merge your changes back in via some other means. On Windows I use WinMerge.

windows-server-2003:
> why does "if" need two args

My description was not enough.(sorry, there were no description) The "if"command is similer to "ifeq" and the condition is matching the two args. I won't add "true" or "false" because adding them has no meaning. For now, the only type of vars in GM9 is String and I think there are no more conditions should be added. I'm thinking to change "if" to "ifeq". 

> control flow in external function

so I can make some changes in ExecuteGM9Script() ? To jump somewere, the external function must get the pointer of "line_end" from somewere.

note:I have final exams in 2 weeks and I can't use my PC. sorry but please leave this PR until I finish them. I can use my tablet and comment to Github.
The last commit is temporary and incomplete. Don't mind.

d0k3:
Yup, I think it's best to let `ExecuteGM9Script()` handle script pointer repositioning via an external function. I added some info on GBAtemp, cause I want Kazuma77's opinion - that's okay, too, right?

windows-server-2003:
GBATemp, yes, we need "if" nesting. Writing scripts without "if" nesting is also complexity. There are some problems for "if" nesting and "for" but possible. I should change the "if"'s condition to like Linux's one.

The script preview return once to the start of the line when start to search a label. It's unnatural so I disabled (don't update) preview while jumping.

d0k3:
@windows-server-2003 - let me know once I should review the code again. Thanks for all the effort you're putting in!

d0k3:
Okay, I had a quick look. Before we continue - make sure you concentrate on your exams, alright?

Now, the label handling - that's still much too complicated. `run_line()` does not even need to know what a label is, and we don't need no max label count.

All you have to do for labels in `run_line()` is change this line:
`if ((cmd >= line_end) || (*cmd == '#')) return true; // empty line or comment`
to this:
`if ((cmd >= line_end) || (*cmd == '#') || (*cmd == '@')) return true; // empty line or comment or label`

Also, `search_label` - I wouldn't allow it to jump right into a conditional (`if`) block. See also here:
https://stackoverflow.com/questions/8481558/windows-batch-goto-within-if-block-behaves-very-strangely
... and it's similarily with other kinds of shells.

It's better to just let goto reset the if nesting lvl to zero, then search for a label, then just assume this label is outside of any if conditions (of course, output an error if we encounter a stray `else` or `end`). Thus, `search_label` could work much simpler (pseudo code here):

for each line starting at the first:
 skip all whitespaces
 if (end_of_line or curr_char != '@') continue
 found if label matches

I think, you could also simplify the conditional skipper code. I'm unsure if that's what `search_command` is for. In any case, on any `if` inside a conditional skipped block, all you have to do is increase the count, and decrease the count if an `end` is encountered. If the count reaches 0, you reached the end of the data you had to skip (hope that's understandable). Of course the handling for an else is a little bit different, but nothing too special. You could do all that stuff in a simple small function.

In fact, a lot less global vars may be required if you forbid `goto` inside a conditional block (which you should!) and just keep the current level inside a global var. Maybe the current level is almost all you need? Do we even need to limit the level of if nesting then?

In general, keep your changes much, much more simple and more contained. You don't have to catch all mistakes script editors may make. If they do fuck up, well, let them fuck up. Catch extraneous `end` and `else`, but for everything else, leave it to the script devs.

windows-server-2003:
I disabled showing error "'end' expected" because I can catch it only if skipping conditional block.

windows-server-2003:
About global vars
>u32 ifcnt

Count of if nesting. increases at 'if's and decreases at 'end's unless you're jumping to a label from 'goto'
>u32 ifcnt_skipped

increases at skipped 'if's and decreases at skipped 'end's and reset to 0 on finishing skipping/starting skipping(for safety, double-reset it)
example : 
```
if chk test test1 # fail
　## 0
　if chk test test
　　## 1
　end
　##0
end # stop skipping
```
It won't be used if we aren't skipping. On 'end' and else', if skipping and it is 0, it will stop skipping.

>u8 skip

0 -> not skipping
1 -> 'if' with failed condition command. From 'if' to 'else' or 'end'
2 -> 'if' with succeed one. From 'else' to 'end'
3 -> searching a label. Set in run_cmd() by 'goto' and used to decide to run search_label() in ExecuteGM9Script().  Reset if found the label.

> bool syntax_error

One to know that a syntax error occurred in run_cmd() and should disable -o and -s in ExecuteGM9Script().
I know this way is not good. so If you have any better ideas to disable -o & -s teach me!

> bool running_cond

True if running a command as a condition of 'if'.
This is to prevent infinite recursive 'if' condition running. infinite recursive("if if if...") makes no sense and may lead to a bug or crash cause memory overflow.

> bool if_cond_res

The result of a command as a condition of 'if'.
set in run_line()(at calling itself) and used in run_cmd()

> char findlabel[513]

The name of finding label.
findlabel[0] == '\000' means not jumping now.



Any opinion to this ?

d0k3:
Dammit. Okay, i fucked up a little, don't worry too much about this being closed now and not merged. I'll review later today!

d0k3:
Alright, I merged and squashed your commits into my if-else branch. Also had a first look already and I think I'll merge (properly, not like my messup a few minutes ago) that today.

windows-server-2003:
Thank you!

Yes, I'm checking it too.

PR #263, opened 2017-11-11
Put the SD init timeout behind a compile flag

### LINKED ISSUE #262: In ntrboot mode, GM9 fails to automatically mount large SD card.
Using the latest commit, with the proper aeskeydb.bin compiled into it.

Most of the (small, ~4GB) SD cards seem to automatically mount fine. However, my main 3DS's 128GB card does not automatically mount when booting GM9 in ntrboot mode. Pressing R+B to remount the card works perfectly fine, however this is not something that you can do when GM9 is compiled with an autorun script.

One other relevant detail is the SD card has 2 EmuNANDs (RedNANDs).

angelsl:
Fixes #262

### DIFF
--- a/source/nand/sdmmc.c
+++ b/source/nand/sdmmc.c
@@ -397,7 +397,9 @@ static int SD_Init()
     u32 temp = (handleSD.error & 0x1) << 0x1E;
 
     u32 temp2 = 0;
-    s32 timeout = 0x80;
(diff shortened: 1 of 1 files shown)


d0k3:
Taken it over, `new_make2` branch, which will make it's way into master soon.

PR #268, opened 2017-11-26
Fixed some bugs.

profi200:
The main bug was not initializing PXI on ARM9. This can break GodMode9 if launched from certain chainloaders.


d0k3:
Merged locally, thank you!

PR #278, opened 2017-12-19
Fix building the tar archive on macOS (and probably BSD)

AuroraWright:
Tested with GNU tar (included in Linux distros) and BSD tar (included in macOS/BSD).
"--disable-copyfile" is macOS specific and causes an error on GNU tar so it's only added if macOS is detected

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ $(if $(wildcard $1),-C"$(shell dirname "$(abspath $(1))")" "$(shell basename $(1
 endef
 
 define ADDTARFOLDER
-$(if $(wildcard $1/*),-C"$(abspath $(VRAM_DATA))" $(shell find $(VRAM_DATA) -mindepth 1 -maxdepth 1 -exec basename -a {} + | sed 's,\(.*\),"\1",'))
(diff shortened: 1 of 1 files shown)


d0k3:
To be honest, I'm unsure. `ustar` format is not a good idea, cause this way you're user and group name will be included in the file. `v7` is the only format that doesn't do this. Also, the `ls` on the tar command is there so that a missing file (like no README.md or no slash) doesn't cause the build to fail. I think the way it is now, you will have a failed build with a missing readme or splash.

d0k3:
More problems... this one will not allow to add dirs from data/ to the tar, however dirs are required. it will just create a "flat" archive.

Is ustar the only format available on Mac? Cause, as I said, this one leaks personal info which many users may not want. Maybe it would be a better idea to create a TAR variable and point that to the one dkA installed? On Windows that is `C:\devkitPro\msys\bin`, and assuming dkA is available is a safe bet. I also think this bloats the Makefile a bit, but this is a minor issue and perfectly solvable.

d0k3:
Okay, thanks a lot for this one. @ihaveamac did provide a Python script that further streamlined this process, thus solving the problem on all systems. I took his pull request (which wouldn't have happened without your work), so I'm closing this now.

PR #279, opened 2017-12-20
Scripting: Added ssplit command

windows-server-2003:
A command to cut string at a char

ssplit [-b | --before] [-f | --first] VAR "string" "c"

Example
~~~
set PATH "0:/path1/path2/file.txt"

ssplit STR $[PATH] "/"
echo $[STR] # file.txt

ssplit -b STR $[PATH] "/"
echo $[STR] # 0:/path1/path2

ssplit -f STR $[PATH] "/"
echo $[STR] # path1/path2/file.txt

ssplit -b -f STR $[PATH] "/"
echo $[STR] # 0:


## errors
ssplit -o STR $[PATH] "b" # not found
echo $[STR] # 0: (keeps the last string)

ssplit -o STR $[PATH] "something" # not a char
echo $[STR] # 0: (same)
~~~

### DIFF
--- a/arm9/source/utils/scripting.c
+++ b/arm9/source/utils/scripting.c
@@ -69,6 +69,7 @@ typedef enum {
     CMD_ID_INPUT,
     CMD_ID_FILESEL,
     CMD_ID_SET,
+    CMD_ID_SSPLIT,
     CMD_ID_CHK,
     CMD_ID_ALLOW,
     CMD_ID_CP,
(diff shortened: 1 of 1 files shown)


d0k3:
There really only is a minor problem with that - you shouldn't write directly to argv[x]. That one I'll fix myself quickly.

d0k3:
Taken over locally as strsplit, thank you!

PR #280, opened 2017-12-22
Add a custom Python script "add2tar.py" to create the vram0 tar file

ihaveamac:
This creates a tar file using the ustar format but doesn't include any extra information (user, group, date/time, etc). This works without depending on OS-level tools with different behavior (e.g. bsd-tar and gnu-tar which are not compatible with the same arguments).

`add2tar.py` requires Python 3.5 or later.


d0k3:
Merged locally. Thanks a lot, also to @AuroraWright, cause her work lead up to this!

PR #305, opened 2018-01-29
load gm9 if no payload is found

eiiiiix:
if your use the bootloader function and you dont have a payload to boot it will normally turn off
but puting that after the BootFirmHandler will load gm9 instead of shuting down
if a payload to boot is found this will not be reached because it stops the function to boot the payload
this is 99.99% safe
i just dont know what happens if a broken payload is loaded because i have no broken payload to test it with
but if a broken payload is used im pretty sure the 3ds will just shutdown

### DIFF
--- a/arm9/source/godmode.c
+++ b/arm9/source/godmode.c
@@ -1934,6 +1934,9 @@ u32 GodMode(int entrypoint) {
         if (IsBootableFirm(firm_in_mem, FIRM_MAX_SIZE)) BootFirm(firm_in_mem, "sdmc:/bootonce.firm");
         for (u32 i = 0; i < sizeof(bootfirm_paths) / sizeof(char*); i++) {
             BootFirmHandler(bootfirm_paths[i], false, (BOOTFIRM_TEMPS >> i) & 0x1);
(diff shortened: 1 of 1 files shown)


d0k3:
Can't take this right now, big refactor in progress. Leaving it open for the time being.

eiiiiix:
alright hope the refactor foesnt mess anything up

SirNapkin1334:
That'll explain the lack of commits (not that that's a bad thing). Well, that's pretty cool! Hope it goes smoothly, and with less build errors than Eix encountered when he started working on EM9 again.

eiiiiix:
@SirNapkin1334 check the hellomalloc branch [https://github.com/d0k3/GodMode9/tree/hellomalloc](url)

d0k3:
Added in https://github.com/d0k3/GodMode9/commit/7235401b2bff25d34cf55634b5e77164d57812b8 - but had to rewrite it cause of the refactor.

I also added a prompt in that case. The bootloader really should be used as such (a bootloader), booting right into GM9 is just a helper construct.

eiiiiix:
alright
but some people may keep all payloads on the sdcard and sometimes boot without it

PR #320, opened 2018-02-27
Forbid writing to SD if write locked

windows-server-2003:
The bit 7 seems to be a SD card write lock flag(0 if locked).

### DIFF
--- a/arm9/source/filesys/fsperm.c
+++ b/arm9/source/filesys/fsperm.c
@@ -26,6 +26,12 @@ bool CheckWritePermissions(const char* path) {
     if ((drvtype & DRV_IMAGE) && !CheckWritePermissions(GetMountPath()))
         return false; // endless loop when mounted file inside image, but not possible
     
+    // SD card write lock check
+    if ((drvtype & (DRV_IMAGE | DRV_SDCARD | DRV_EMUNAND | DRV_ALIAS)) && SD_WRITE_LOCKED) {
+        ShowPrompt(false, "SD card is write locked.\nCan't continue.");
+        return false;
+    }
+    
     // check drive type, get permission type
     if (drvtype & DRV_SYSNAND) {
         u32 perms[] = { PERM_SYS_LVL0, PERM_SYS_LVL1, PERM_SYS_LVL2, PERM_SYS_LVL3 };
--- a/arm9/source/filesys/fsperm.c
+++ b/arm9/source/filesys/fsperm.c
@@ -4,6 +4,7 @@
 #include "image.h"
 #include "unittype.h"
 #include "ui.h"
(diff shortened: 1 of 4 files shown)


SirNapkin1334:
Interesting! I hadn't realized that you could write to the SD if it was locked, though it does make sense since the switch doesn't actually set anything on the inside of the SD, and there's no reason for the 3DS to include that functionality since it always has to write to the SD. But, apparently it does detect it, and ignore it? I can test it in about 9 hours.

SirNapkin1334:
I checked it just now on my o3DSXL, and it worked. But you know this will be entirely useless on n3/2DS(XL), since microSD doesn't have a lock unless inserted into an adapter.
Good job, though!

SirNapkin1334:
And, it functions as expected on the new systems (tested on n2DSXL), always allowing writes.

arm9/source/filesys/fsperm.c
```diff
@@ -26,6 +26,12 @@ bool CheckWritePermissions(const char* path) {
     if ((drvtype & DRV_IMAGE) && !CheckWritePermissions(GetMountPath()))
         return false; // endless loop when mounted file inside image, but not possible
     
+    // SD card write lock check
+    if ((drvtype & (DRV_IMAGE | DRV_SDCARD | DRV_EMUNAND | DRV_ALIAS)) && SD_WRITE_LOCKED) {
```
d0k3: Images can already have their origin on CTRNAND. You need to check the drive type of the image source. Besides, you don't need to check for images here. See the instrcuction block right above?

arm9/source/filesys/fsperm.h
```diff
@@ -27,6 +27,8 @@
 #define PERM_YELLOW     (GetWritePermissions()&((PERM_SYS_LVL1&~PERM_SYS_LVL0)|(PERM_EMU_LVL1&~PERM_EMU_LVL0)|(PERM_SDDATA&~PERM_SDCARD)))
 #define PERM_GREEN      (GetWritePermissions()&(PERM_SDCARD|PERM_IMAGE|PERM_RAMDRIVE|PERM_EMU_LVL0|PERM_SYS_LVL0))
 
+#define SD_WRITE_LOCKED (!((*(volatile u16*)0x1000601C) & (0x1<<7)))
```
d0k3: That really does not belong in here if it's checking an SD card specific register. It most likely belongs into sdmmc.h.

d0k3:
@windows-server-2003 - just change the two things I remarked in the review, and I'll merge it.

windows-server-2003:
I'll do it in 5 hours.

d0k3:
Merged locally, thank you!

PR #337, opened 2018-04-03
Add April Fools compiling

SirNapkin1334:
I thought that it would be nice to be able to compile with (or without, that'd be helpful for some people) the April Fools features. To compile *with* April Fools, simply run `make APRIL_FOOLS=xxxx` where `xxxx` is the year of the prank (currently the only available year is 2018 for obvious reasons). To compile without, simply run `make`. This shouldn't require a ton of maintenance, though it should be noted that from now on  that the VRAM0 support file naming (i.e. `VRAM0_FONT_PBM`) in the code must always be replaced with their LOTTERY equivalents (i.e. `LOTTERY_FONT`) (if you really need to be able to use VRAM0 naming than I'll see if I can figure out how to do this - this is my first C coding project, maybe my next will be doing the aforementioned). I tested compiling with and without APRIL_FOOLS and they both worked, so I would consider this a success!


SirNapkin1334:
Shoot, I realized that this should be merged into d0k3:master so it doesn't just sit there and not be used, if possible please merge it to master!

d0k3:
Yeah, it's not in master, and it's not in there for a good reason. The lottery thing was a quick shot, and the implementation could be a lot better. I do understand that people actually like the bricked modes, and I might do something about that. Making the April Fools code permanent is not the right solution, though, and would hinder future progress.

PR #338, opened 2018-04-04
Fix 'sd is write protected' after sd ejection

windows-server-2003:
Fixes showing 'SD is write protected' when trying to write something to SD after SD ejection without unmounting.(It should just fail)
Also fixes some typos.


windows-server-2003:
It checks if the 5th bit is on and the 7th bit is off.
The 7th bit seems to be off if SD is missing or write protected and otherwise, on.
But we should not show write protection error if SD is missing(of course, trying to write to SD without SD is unexpected though)

And yes, it catches correctly.

d0k3:
Alright, merged, thank you!

PR #343, opened 2018-04-11
Add support for IPS/BPS/BPM-formatted binary delta patches.

Hyarion-SS:
This is a straightforward C port of the "beat patch" application code from byuu's beat v03, adapted to GodMode9's virtual FatFS. GM9 permissions checking, progress bar, and nominal scripting support were also added. Files are currently read directly instead of reading into buffers - this slows down the process, but allows theoretical use of any size of patch and is a good base for optimization where possible.

While I started this for our own use, I believe that it has great potential for the rest of the 3DS modding community, especially when coupled with the standalone scriptrunner build option for GM9. Modders can release legal delta patches in a well-known standard packed directly into a standalone GM9 firm - the autorun.gm9 script inside will then locate the original game files and seamlessly patch them into the correct location for use with LayeredFS with a single button press. 

As a disclaimer, I'm an engineer, not a programmer. It's been a decade since I touched C and I may have made some odd choices as a result. Regardless, this follows the original as closely as possible and is intended more as a a mostly tested and fully working, if sometimes slow, implementation that can be polished and optimized for the 3DS. Even its current state, however, it's fast enough for all but the largest mods. Optimizing reads and writes and using the RAM where possible are next on the list, perhaps copying the files into the ramdrive before patching if the space is available, but I'm unfamiliar with the GM9 ecosystem and would definitely appreciate any help.

Thanks for this great tool, by the way.


d0k3:
Okay, that came out of nowhere. But, the implementation looks pretty nice and clean and I see the potential of this. I personally still prefer IPS over BPS, but, ofc, I also see the downsides of IPS.

I'll still have some reviewing to do, maybe I'll have some nitpicks and ideas for improvement. Two minor nitpicks that catched my glance so far:
* I'd hide more of that stuff (defines, enums) inside the .c file. In fact, I'd only expose what's actually used outside of `bps.c`. Personal preference here.
* I wouldn't include `beat/beat.h` but rather edit the `Makefile` to include the `beat` path and then just include `beat.h`
Very minor, I know.

As for using the ram drive - I'd rather not have some files pop up there unexpectedly (for the user, that is). Script writers can choose to use the ram drive, I'd rather not have this by default. Is the heap not enough?

As for speeding this up... I may be able to help if you pinpoint the locations where there is a lot of time spent. Reading files without a buffer, f.e., is slow with the 3DS hardware. Using the heap and STD_BUFFER_SIZE is prefered.

SirNapkin1334:
A question - why did you implement what seems to be CRC32 code? Would this (down the line) also allow us to do CRC32 checks along with SHA256 checks?

Hyarion-SS:
Nice, thanks. I have no issue with IPS - it just didn't meet our needs. It would be very simple to add an IPS patcher function in case other users would want it, though.

I'll adapt it to your coding practices, no problem. I wasn't aware of STD_BUFFER_SIZE - knowing that makes heap usage much easier. However, one of the files we need to patch is larger than 1MB, and initial indications are that reading from the source file is the largest bottleneck. Holding only 1MB of the source file at a time in memory is unfortunately not an option, as the entire length needs to be accessible due to the nature of BPS. If I may suggest, I could implement an optional flag -r:

Default behavior: If source file is <1MB, use a buffer. Otherwise, display a prompt - "patching this file will likely be slow, do you wish to use the ramdrive?" If accepted, switch to -r behavior. If denied, fall back to direct reads.
-r behavior: Use the heap for <1MB source files, but instead attempt to copy any >1MB source file to ramdrive, deleting it after patching. Fall back to direct reads if the copy failed.

The -r flag can then be used in scripts like ours where the end user will never be looking at the ramdrive and probably does not even know what it is. As it's a standalone script runner without file manager functionality, there should be no downside.

@SirNapkin1334 - CRC32 is baked into the BPS format as a simple sanity check that the patch is intact, that it is being applied to the intended source, and that the intended patched file has been produced. The code to generate a CRC32 checksum on any file exists, but it isn't terribly useful in most cases.

d0k3:
@SirNapkin1334 - of course we could add CRC32 support. We could also add CRC16 support, SHA224, MD5, etc... but, would it make sense? I'd say no, it would only make things more confusing to the user. Better to stick with one checksum.

@Hyarion-SS - You know, the heap is bigger than just those 1MB - I think it's 32MB, unsure. Now, I'm pretty sure you're aware, but even the ramdrive space is limited, and, even worse, the limit depends on wether you use an O3DS or N3DS. That means, if you go for the ramdrive solution, you haven't solved the problem, you just postponed it.

So, still not liking the ramdrive solution. As I said, script writers can choose to do the copy, they are also more likely to know which sizes to expect. You could use the heap (of course, for the love of god, check the output of `malloc` :)), you'd hit the limit only a little faster than with the ramdrive.

Alternative to that, but pretty ugly to implement: use a 'floating window' buffering scheme, where you have a standard sized buffer that is 'moved' along the file. You'd need to keep track of what is in the buffer and what is not, so there is potential for error.

Hyarion-SS:
You're right, I was too hesitant to malloc. I still appreciate having the direct read fallback just in case, but it's blazingly fast now in comparison.

SirNapkin1334:
Actually, md5 could be useful since many people use it as a quick and easy that's available on all platforms. In the README you give the md5 of the aeskeydb, so why not implement it?

Hyarion-SS:
As it turns out, we do have use for IPS as well. I'll drop that in a separate PR whenever I get around to implementing it.

d0k3:
Alright, so I'm assuming you're happy with the current state of the pr, it is tested, and ready to merge, right? Still haven't done a proper review, but I'll get to that quick.

d0k3:
Great! Give me a day or two so I can look though the code. Doubt I'll find another nitpick, tho.

d0k3:
Merged, thanks a lot!

PR #355, opened 2018-04-22
Add scripting commands extrcode, cmprcode, cp -p

Hyarion-SS:
* extrcode already existed, but now has a progress bar and an entry in HelloScript.
* cmprcode recompresses a code binary into its original format. It takes 2-3 minutes.
* cp -p appends file 2 to the end of file 1 rather than overwriting it.
* cp displays a full progress bar instead of % completion if the file to be copied is less than the buffer size.


SirNapkin1334:
I like the idea, especially of cp -p since things like CTRtransfers and ticket copies because it looks like they’ll be quick at first but then you can’t judge how long it’ll take because there are just more and more progress bars. +1

arm9/source/game/codelzss.c
```diff
@@ -91,3 +97,276 @@ u32 DecompressCodeLzss(u8* code, u32* code_size, u32 max_size) {
...
+            memcpy(a_pCompressed, a_pUncompressed, uOrigSafe);
+            memmove(a_pCompressed + uOrigSafe, pCompressBuffer + uCompressSafe, uCompressedSize);
+            memset(a_pCompressed + uPadOffset, 0xFF, uCompFooterOffset - uPadOffset);
+            CodeLzssFooter* pCompFooter = (CodeLzssFooter*)(a_pCompressed + uCompFooterOffset);
+            pCompFooter->off_size_comp = uTop | (uBottom << 24);
+            pCompFooter->addsize_dec = a_uUncompressedSize - *a_uCompressedSize;
+        }
+    }
+    return bResult;
+}
```
d0k3: Newline missing.

arm9/source/game/codelzss.c
```diff
@@ -44,6 +46,10 @@ u32 DecompressCodeLzss(u8* code, u32* code_size, u32 max_size) {
     
     // main decompression loop
     while ((ptr_in > comp_start) && (ptr_out > comp_start)) {
+        if ((counter++ == 0) &&
+            !ShowProgress(data_end - ptr_out, data_end - data_start, "Decompressing .code...") &&
```
d0k3: Given that this typically takes almost no time at all (you may convince me otherwise), does it really make sense? This is the only place the progress bar is used and also the only reason for the `ui.h` dependency (which makes the code less portable). Also keep in mind that the progress bar has to be initialized / reinitialized with `current == 0`.

EDIT: yeah, I see, the prgress bar is used in that other function, too. In general, only introduce dependencies like `ui.h` it it is really required. This one is most likely justified for the compression function.

arm9/source/game/codelzss.c
```diff
@@ -91,3 +97,276 @@ u32 DecompressCodeLzss(u8* code, u32* code_size, u32 max_size) {
     *code_size = data_end - data_start;
     return 0;
 }
+
```
d0k3: Starting here: That code could use a cleanup. More consistency in function names (use lower case for all 'private' functions for example), and no newlines before `{`. Part of this is pretty hard to read, too. I'm also wondering why that has to take that long. I do understand that LZSS is an asymmetric compression algo, but the requirement here seems a bit harsh.

arm9/source/utils/gameutil.c
```diff
@@ -1584,6 +1584,53 @@ u32 ExtractCodeFromCxiFile(const char* path, const char* path_out, char* extstr)
...
+    if (!path_out && (fvx_rmkdir(OUTPUT_PATH) != FR_OK)) return 1;
+    strncpy(dest, path_out ? path_out : OUTPUT_PATH, 256);
+    if (!CheckWritePermissions(dest)) return 1;
+    
+    // allocate memory
+    u32 code_dec_size = fvx_size(&code_file);
+    u8* code_dec = (u8*) malloc(code_dec_size);
+    u32 code_cmp_size = code_dec_size;
+    u8* code_cmp = (u8*) malloc(code_cmp_size);
+    if (!code_dec) {
```
d0k3: Not checking `code_cmp` here, also not freeing stuff in case only one got not allocated. Isn't it possible to do that in only one buffer, same as decompression? You will hit memory limits fast this way.

arm9/source/utils/scripting.c
```diff
@@ -526,6 +528,7 @@ u32 get_flag(char* str, u32 len, char* err_str) {
     if ((len < 2) || (*str != '-')) flag_char = '\0';
     else if (len == 2) flag_char = str[1];
     else if (strncmp(str, "--all", len) == 0) flag_char = 'a';
+    else if (strncmp(str, "--append", len) == 0) flag_char = 'a';
```
d0k3: I hope you made sure there's absolutely no way for `--all` and `--append` to clash. Up to this point I made sure everything uses a letter not used elsewhere.

d0k3:
@Hyarion-SS - sorry, don't want to nag, but this has been open with some requested changes for a while. Are you still working on it or intend to?

Hyarion-SS:
Yes, I will be returning to it this weekend -- had to set it aside for a couple of weeks as I'm an engineering grad student.

ParzivalWolfram:
Nice, looking forward to the new commands! I'd kinda like to see BATCH-style logic added ("if exists" and such) to help make smarter scripts easier to make.

MadScript77:
Um, "if exist" is already implemented.  So is "not" for that matter.  Meaning you can also do an "if not exist" -- which is often even more convenient.  You can do an "if not exist $[FILENAME]" followed by an "echo" saying the needed file wasn't found, then a "reboot" followed by an "end" command.  Or maybe check for "not $[ONTYPE]" and follow it with "goto" and "end" commands, as opposed to having half of the entire algorithm under the if conditional (the "elif" part being a continuation thereof, of course).  This does require an extra "goto" to skip the section that is being jumped to.  Still, probably less messy overall.

Most of my scripts would probably have more "goto" and less "elif" statements had I written them AFTER "not" was implemented.  It would be too much trouble to change them now, though.  Things would break left and right because some of the needed changes would likely be missed initially.

arm9/source/filesys/fsutil.c
```diff
@@ -507,56 +513,61 @@ bool PathMoveCopyRec(char* dest, char* orig, u32* flags, bool move, u8* buffer,
...
+            if (calcsha)
                 sha_update(buffer, bytes_read);
         }
         ShowProgress(1, 1, orig);
         
         fvx_close(&ofile);
         fvx_close(&dfile);
-        if (!ret) fvx_unlink(dest);
-        else if (!to_virtual && flags && (*flags & CALC_SHA)) {
+        if (!ret && ((dsize == 0) || (fvx_lseek(&dfile, dsize) != FR_OK) || (f_truncate(&dfile) != FR_OK))) fvx_unlink(dest);
```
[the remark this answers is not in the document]
Hyarion-SS: Extensively.

arm9/source/game/codelzss.h
```diff
@@ -6,3 +6,4 @@
 
 u32 GetCodeLzssUncompressedSize(void* footer, u32 comp_size);
 u32 DecompressCodeLzss(u8* code, u32* code_size, u32 max_size);
+bool CompressCodeLzss(const u8* a_pUncompressed, u32 a_uUncompressedSize, u8* a_pCompressed, u32* a_uCompressedSize);
```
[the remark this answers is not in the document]
Hyarion-SS: Added.

arm9/source/utils/scripting.c
Hyarion-SS: Copy does not use --all, only --append.

arm9/source/filesys/fsutil.c
```diff
@@ -452,10 +455,13 @@ bool PathMoveCopyRec(char* dest, char* orig, u32* flags, bool move, u8* buffer,
...
     
+    u64 current = (progress) ? 0 : 1;
+    u64 total = (progress) ? 0 : 1;
+    
     // the copy process takes place here
-    if (!ShowProgress(0, 0, orig) && !(flags && (*flags & NO_CANCEL))) {
+    if (!ShowProgress(current, total, orig) && !(flags && (*flags & NO_CANCEL))) {
         if (ShowPrompt(true, "%s\nB button detected. Cancel?", deststr)) return false;
-        ShowProgress(0, 0, orig);
+        ShowProgress(current, total, orig);
```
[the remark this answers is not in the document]
d0k3: Alright, this should work.

d0k3:
Okay, good. We've got one big problem left.... I can't just merge that without fucking up the commit history completely. Sooo.... could you maybe redo the commits on the current master, using something like WinMerge (or equivalent)?

You don't have to do the exact commits, just something that makes sense.

Hyarion-SS:
I rebased this branch onto the current revision of master. That should keep everything tidy, but let me know if you want any more housekeeping.

d0k3:
Thanks, this looks great! Looks like your better with that git stuff than me. I'll have a look later today. Doubt there will be more nitpicking from my side, but I'll get back.

d0k3:
Okay, there's some stuff I still want to discuss and read your opinion on...

How does `cp -a` handle a folder copy? Trying to copy a file to a folder or vice versa should already fail gracefully, but I'm unsure what happens when a user tries to `cp -a` from folder to folder. It would be somewhat bad (I think, I'm not sure) if in that case all the files inside an existing folder would be appended to. Maybe it would be a good idea to check if the input is really a file in case of `-a`.

Shouldn't `cp -p` just be the default for small files? As I understand, `-p` is intended to make the progress bar less messy (displaying always full) if a lot of small files are copied in succession (I may be wrong). We could also just make this behaviour the default for folder names and for files sized smaller than the buffer size (currently 1 MiB).

Hyarion-SS:
We've gone through another round of scriptrunner testing on this codebase, and it looks good.

d0k3:
Merged locally, with some small changes as discussed. Thank you!

PR #356, opened 2018-04-22
Switch iteration order over RomFS's to "files before folders".

Hyarion-SS:
* This should only affect the For loop in scripts, causing files inside of a RomFS to be read in the order described on 3DBrew: https://www.3dbrew.org/wiki/RomFS#Directory_Metadata_Structure

* Was previously "folders before files", but the For handler starts reading directory contents immediately after encountering them.


d0k3:
Okay, merged locally. While this commit is completely under-the-hood and won't have a noticable effect, it will make a difference for `for` recursion in that files will come before folders.

As I stated earlier, I don't like the reason behind this commit, because it is a hacky solution for what's intended to be achieved and may be changed later in the GM9 code, causing regressions. Finding a better solution is encouraged, for now this will do.

PR #362, opened 2018-04-26
Replaced the old non-branching NAND scripts.

MadScript77:
I replaced the old non-branching NAND backup/restore scripts with a branching, hotkeyed one.


SirNapkin1334:
Personally, I think that it should have a prompt saying something like this:
```
!!! WARNING !!!
THIS SCRIPT WILL COMPLETELY RESTORE YOUR NAND.
THIS MAY OVERWRITE YOUR SIGHAX INSTALLATION.
YOU HAVE BEEN WARNED.
```
Instead of checking that you're using ntrboot. If you want a better, but more complicated solution, mount the NAND backup and check if firm0/1 match the FIRM in CTRNAND. If they match, warn the user that it will overwrite their B9S installation. If they don't, warn them that if the image is corrupt it will likely brick their system. Make sure to inclue the EmuNAND warn. Besides that, I like it. I can't wait to use it!

MadScript77:
I just reverted it to what d0k3 originally wanted.  The check is commented out in InScripted's version.  I thought it did say something to that effect, but, I can change it for clarity (or d0k3 can, of course).  You don't have to wait to use it.  It's in my fork.  Grab it, run it, see what you think.

I doubt there are many corrupt EmuNANDS that have GM9-style names in the first place, but I can add partial SHA checking if there is a need (partial SHA, 0x200/0x400 bytes starting at 0x200 or something).  Though for the moment I'm kind of busy replacing all instances of "labelsel" with "keysel" in my scripts.  Not to mention putting together a few custom chainloaders.  The next release will have 3 -- Easy Mode, Kids Mode, and an advanced, fully configurable beast with many hotkeys (all but two being user-configurable -- Settings needs to always be there, and I reserved "down" for a "keysel" GUI menu for people that can't remember hotkeys).  I'm heavily leaning towards giving it the name of the AIO itself.

windows-server-2003:
Nice.

I found that the 7th option in the main menu(<RIGHT> Transfer SysNAND to EmuNAND) is not fully displayed though.

Also I can't exit without shutting down or rebooting.

SirNapkin1334:
Yes, I haven't had a chance to test it, since I've been at school today, but I'll take a look at it as soon as I can. (1-2 hours from now)
I anticipate checking it out!

MadScript77:
Well, I use most of my scripts in compiled script runners, so, I tend to loop stuff that's not for the swap card.  I can add an exit option.

SirNapkin1334:
I believe that the poweroff/reboot/exit buttons are kinda off. START should be reboot, since that's what it normally is in GM9. It doesn't matter that much to me as much which which of the remaining two is SELECT or DOWN, but I wholeheartedly believe that START should be reboot.

MadScript77:
Hmm, the current iteration of my AIO would drive you nuts then, probably.  I just kind of went A, X, Y, L, R, left, right, Select, Start for most stuff.  I'm actually using UP, not DOWN -- too many people will be used to tapping down to navigate.  I only do that with a two-screen menu like InScripted/InScripted_GUI has (one's the full chainloader, the other's just the on-screen menu part of it).

d0k3:
Just tell me when it's ready to merge :)

MadScript77:
Well, I think it's ready, @d0k3, unless you can think of a way to check for a firm mismatch without it triggering on hacked NAND images.

d0k3:
Merged locally, thank you!

PR #369, opened 2018-05-11
fix format vulnerability

windows-server-2003:
* Fix some vulnerability caused by passing char pointer directly to format argument
* Added warning option for that


windows-server-2003:
Added some more the same type fixes(modified commit)

PR #374, opened 2018-05-23
Better readme fix

SirNapkin1334:
Thanks @ihaveamac for pointing out why the readme fix was put in, didn't realize markdown did that (also didn't notice that section of the README). This is a solution that simply escapes that, so you can see the buttons and they're like that of inside GodMode9, rather than using []. Also more similar to how The Guide is written. I also added that when building in SALTMODE the keycombo is START.


d0k3:
Taken over locally

PR #383, opened 2018-06-05
fix buffer overflow

windows-server-2003:
In DrawTopBar() the font width can be 1 at minimum so ```len_path / FONT_WIDTH_EXT``` can be up to 280.
The same for the others.


d0k3:
Okay, the problem with this pull request is, you're grabbing somewhat large chunks ofs memory for a hypothetical scenario. The hypothetical scenario being font widths (including one pixel space) below 4px. It is basically impossible to make letters less wide *and* readable, at least without the use of color.

Sooo... why not use the FONT_WIDTH_EXT macro to make the arrays a reasonable size? You could also limit the size of the strings themselves, of course (clamp it at a certain length, f.e. Or, we could decide not to accept fonts < 4px in width.

windows-server-2003:
Alright, fixed.

d0k3:
Taken over locally, thank you!

PR #388, opened 2018-06-17
Buffer all BPS, IPS, CRC32 reads in chunks of 1 MB or less

Hyarion-SS:
As the title states, all reads are now buffered in large chunks of STD_BUFFER_SIZE. CRC32 was trivial and IPS was straightforward - the nature of the format limits IPS to fairly small files that are read linearly from beginning to end.

BPS, however, necessitated a complete rewrite. Files can be any size and read from any point in the source or target files as many times as they wish, and they take full advantage of this - BPS patches are essentially a compressed archive containing the target file compressed using both the source file and previous sections of the target file itself as a dictionary, with equal or even superior efficiency in my tests to the bzip2-leveraging bsdiff format.

The previous BPS implementation tried to read patch, source, and target files into memory before falling back to simply reading the requested portions of the file on demand. For files of any reasonable size, this resulted in a **ludicrous** amount of tiny reads and writes, usually overlapping with data that had been previously read. The overhead and inefficiency could result in enough slowdown as to be practically unusable, so a growing amount of cruft was devoted to workarounds avoiding this fallback at all costs. I wasn't too happy with it.

So naturally, I wrote my own memory management system.

Instead of an array of bytes, file data is represented by an array of DataChunk structs, themselves each representing non-overlapping STD_BUFFER_SIZEd pieces of the file. Data reads intelligently switch between chunks midstream with a minimum of overhead, chunks are allocated only when requested, and when more memory is required, only the least recently used chunks are deallocated until the needed memory is available.

It's all been tested across a good number of Old and New 3DS's via an installer script patching a few hundred files each go, and we haven't found any issues. Theoretically it should be able to apply any valid BPS/BPM patch now, all at a very reasonable speed - the script previously took at least 20 minutes to complete but now finishes in less than 5.

### DIFF
--- a/arm9/source/crypto/crc32.c
+++ b/arm9/source/crypto/crc32.c
@@ -5,8 +5,8 @@
 #include "crc32.h"
 #include "common.h"
 
-uint32_t crc32_adjust(uint32_t crc32, uint8_t input) {
-    static const uint32_t crc32_table[256] = {
+u32 crc32_adjust(u32 crc32, u8 input) {
+    static const u32 crc32_table[256] = {
         0x00000000, 0x77073096, 0xee0e612c, 0x990951ba, 0x076dc419, 0x706af48f,
         0xe963a535, 0x9e6495a3, 0x0edb8832, 0x79dcb8a4, 0xe0d5e91e, 0x97d2d988,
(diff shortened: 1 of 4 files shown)


d0k3:
Merged, thank you! When measuring time, keep in mind that GM9 now also has an updated SD driver, which is responsible for part of the added performance (for big file operations, I estimate the new SD driver to be ~10% faster).

Issue #395, opened 2018-07-06
DLC is dumped incorrectly


Steveice10:
DLC TMDs are supposed to contain information on *all* DLC contents, regardless of which are actually available on the device. However, GodMode9 strips out the information of any content files not present on the system when dumping a CIA. The content index in the CIA header should instead be used to indicate which contents are present in the CIA file.

Steveice10:
Related PR: https://github.com/d0k3/GodMode9/pull/396

PR #396, opened 2018-07-06
Fix dumping DLC to CIA.

Steveice10:
Instead of modifying the TMD content chunks, just remove missing contents from the CIA header's content index.

### DIFF
--- a/arm9/source/utils/gameutil.c
+++ b/arm9/source/utils/gameutil.c
@@ -1227,24 +1227,22 @@ u32 BuildCiaFromTmdFileBuffered(const char* path_tmd, const char* path_cia, bool
     if (!name_content) return 1; // will not happen
     name_content++;
     
+    u8 present[TMD_MAX_CONTENTS];
+    memset(present, 1, sizeof(present));
+
     // DLC? Check for missing contents first!
-    if (dlc) for (u32 i = 0; (i < content_count) && (i < TMD_MAX_CONTENTS);) {
+    if (dlc) for (u32 i = 0; (i < content_count) && (i < TMD_MAX_CONTENTS); i++) {
         FILINFO fno;
         TmdContentChunk* chunk = &(content_list[i]);
         snprintf(name_content, 256 - (name_content - path_content),
             (cdn) ? "%08lx" : (dlc && !cdn) ? "00000000/%08lx.app" : "%08lx.app", getbe32(chunk->id));
-        if ((fvx_stat(path_content, &fno) != FR_OK) || (fno.fsize != (u32) getbe64(chunk->size)))
-            memmove(chunk, chunk + 1, ((--content_count) - i) * sizeof(TmdContentChunk));
-        else i++;
-    }
-    if (!content_count) return 1;
-    if (content_count < (u16) getbe16(tmd->content_count)) {
-        if (!ShowPrompt(true, "ID %016llX\nIncomplete DLC (%u missing)\nContinue?",
-            getbe64(title_id), getbe16(tmd->content_count) - content_count)) return 1;
-        tmd->content_count[0] = (content_count >> 8) & 0xFF;
-        tmd->content_count[1] = content_count & 0xFF;
-        memcpy(tmd->contentinfo[0].cmd_count, tmd->content_count, 2);
-        if (FixCiaHeaderForTmd(&(cia->header), tmd) != 0) return 1;
+        if ((fvx_stat(path_content, &fno) != FR_OK) || (fno.fsize != (u32) getbe64(chunk->size))) {
+            present[i] = 0;
+
+            u16 index = getbe16(chunk->index);
+            cia->header.size_content -= getbe64(chunk->size);
+            cia->header.content_index[index/8] &= ~(1 << (7-(index%8)));
+        }
     }
     
     // insert contents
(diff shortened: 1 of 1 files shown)


d0k3:
Alright, I'll have a look tomorrow and then I'll merge it. Thanks for this!

d0k3:
Okay... putting this up for discussion. There was a prompt earlier when DLCs are incomplete. Your pull request removed that prompt. Think we should leave it out? Given that DLCs are basically always incomplete, the prompt may be redundant. On the other hand, users may think that their DLCs are complete now (cause no prompt), which they aren't.

Steveice10:
If you want the prompt to be there I can add it back. I just kind of figured users would understand it can only dump what they have.

d0k3:
Nah, I'd rather just ask for your opinion. I think we can leave it as is.

That pull request is also tested?

Steveice10:
Yes. I tested it by dumping all of my DLC, testing all of my DLC in citra with the DLC-fixing pull request, and reinstalling one of my game's DLC on my system using a dumped CIA.

arm9/source/utils/gameutil.c
```diff
@@ -1227,24 +1227,23 @@ u32 BuildCiaFromTmdFileBuffered(const char* path_tmd, const char* path_cia, bool
...
-        memcpy(tmd->contentinfo[0].cmd_count, tmd->content_count, 2);
-        if (FixCiaHeaderForTmd(&(cia->header), tmd) != 0) return 1;
+        if ((fvx_stat(path_content, &fno) != FR_OK) || (fno.fsize != (u32) getbe64(chunk->size))) {
+            present[i] = false;
+
+            u16 index = getbe16(chunk->index);
+            cia->header.size_content -= getbe64(chunk->size);
+            cia->header.content_index[index/8] &= ~(1 << (7-(index%8)));
+        }
+        i++;
```
d0k3: This one can also be included in the `for` line now.

arm9/source/utils/gameutil.c
```diff
@@ -1227,24 +1227,23 @@ u32 BuildCiaFromTmdFileBuffered(const char* path_tmd, const char* path_cia, bool
...
     if (dlc) for (u32 i = 0; (i < content_count) && (i < TMD_MAX_CONTENTS);) {
         FILINFO fno;
         TmdContentChunk* chunk = &(content_list[i]);
         snprintf(name_content, 256 - (name_content - path_content),
             (cdn) ? "%08lx" : (dlc && !cdn) ? "00000000/%08lx.app" : "%08lx.app", getbe32(chunk->id));
-        if ((fvx_stat(path_content, &fno) != FR_OK) || (fno.fsize != (u32) getbe64(chunk->size)))
-            memmove(chunk, chunk + 1, ((--content_count) - i) * sizeof(TmdContentChunk));
-        else i++;
-    }
-    if (!content_count) return 1;
```
d0k3: This check may still make sense and shouldn't be removed.

d0k3:
I think at least the check should go back in. The other thing is nitpicking :^).

arm9/source/utils/gameutil.c
Steveice10: The check also happens near the top of the function; the difference is now, content_count isn't modified, so it doesn't need to be repeated.

Steveice10:
Comments fixed/responded to.

arm9/source/utils/gameutil.c
```diff
@@ -1227,24 +1227,22 @@ u32 BuildCiaFromTmdFileBuffered(const char* path_tmd, const char* path_cia, bool
     if (!name_content) return 1; // will not happen
     name_content++;
     
+    bool present[TMD_MAX_CONTENTS];
+    memset(present, true, sizeof(present));
```
d0k3: Is `bool` guaranteed to be 1 byte here? Granted, this will work even if it's not.

d0k3: You're not sure, either, correct? Maybe `bool present[TMD_MAX_CONTENTS] = { true }` would be more sane?

Steveice10: It's implementation-defined, although I believe it is 1 byte in the case of devkitARM's environment.

Steveice10: If you want, I could flip it to "missing" and initialize it to all zeros instead of explicitly using "true".

d0k3: Okay, suggestion. Use an u8 array instead. Just to be safe. I know, this is again nitpicking, but I don't like the memset when we're not really sure about the size of the elements in the `present` array. Open for discussion.

Steveice10:
Changed to u8 array.

d0k3:
Merged, thanks! That was quick, also sorry about the nitpicking.

This also fixes #395

jroweboy:
@d0k3 Sorry to pester you about this, but could we see a release with this fix soon? This will be required for DLC support in citra (for those people that don't all the DLC for the game) and when we announce that DLC is working, we'll be telling everyone to redump the dlc through godmode9 and that they should update godmode9 first :)

d0k3:
@jroweboy - yeah, I intended to do one anyways. It will still take me a few days to wrap everything up and test ist.

PR #401, opened 2018-07-23
Fixed a typo

DMSalesman:


DMSalesman:
Makes sense. Just found this one while testing the injection feature, though.

d0k3:
Merged locally (I accidentially included it in my latest commit). Commit doesn't list you as author (another accident). I hope that's okay.

PR #402, opened 2018-07-23
Added refined user guide

DMSalesman:
This commit adds an edited README.md to be built into GM9 as user guide.

The scrubbed parts include refs to the GM9 banner, unused MarkDown and
the whole "How to build this / developer info" block.

Links have been replaced with their corresponding anchor text, when
replacing them could be done without altering the sentences' original
meaning.

Certain parts have been further altered, such as the "License"
paragraph, which now includes the full link to GM9's "license.txt".

### DIFF
--- a/data/README.md
+++ b/data/README.md
@@ -84,7 +84,7 @@ With the possibilites GodMode9 provides, not everything may be obvious at first 
 
 ### Basic functionality
 * Manage files on all your data storages: You wouldn't have expected this, right? Included are all standard file operations such as copy, delete, rename files and create folders. Use the L button to mark multiple files and apply file operations to multiple files at once.
-* Make screenshots: Press R+L anywhere. Screenshots are in BMP format.
+* Make screenshots: Press R+L anywhere. Screenshots are stored in PNG format.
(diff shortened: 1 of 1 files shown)


DMSalesman:
@d0k3 In regards to the tool, no, you don't need to explicitly open and edit the ReadMe, but only to maintain the `patch.json.gz` if the tool complains about a pattern it can't find.

I'm planning to release the script soon, so that you can have a look at it. I'm also available for the maintenance of the patch database if need be, so that you'd only need to clone the repo and launch the tool.

d0k3:
Alright. As for this pull request - I'll need to introduce some changes to the readme myself. Upcoming release. I suggest we leave it open so you can update after I'm done.

DMSalesman:
Expected that, what with the new trimmer and all; no problem.

d0k3:
Alright, I made some small changes to the readme. Could you update accordingly?

Btw, you could actually make Unmark more generic, ie. do it in a way that allows using it for other Markdown files, too. Maybe there's some demand for that.

d0k3:
Merged via the command line, thank you!

PR #404, opened 2018-07-30
Update NANDManager.gm9

TheGinGear:
Just a quick fix, replaced the old keysel command


d0k3:
Yeah, replacing the `keysel`command was bound to bring some problems. Thanks for the quick fix!

PR #405, opened 2018-07-30
Improved file selector

windows-server-2003:
- Show "(dir)" on directory entry.
 - Allow scrolling in select menu.


d0k3:
I'm a little bit unsure about this, cause we still have a limit for the number of displayable entries. I suggest not to increase the limit for now, though. Will have a more thorough look later, looks good so far.

d0k3:
Ah, you got me wrong. I meant the number of entries in the buffer, and it's good as it is now. Using 10 for displayable entries is fine, too.

arm9/source/filesys/fsutil.c
```diff
@@ -818,8 +819,8 @@ bool FileSelectorWorker(char* result, const char* text, const char* path, const
                     char* dot = strrchr(temp_str, '.');
                     if (dot) *dot = '\0';
                 }
-                ResizeString(opt_names[n_opt], temp_str, 32 - 6, 8, false);
-                if (entry->type == T_DIR) snprintf(opt_names[n_opt] + 32 - 6, 6 + 1, " (dir)");
+                ResizeString(opt_names[n_opt], temp_str, _MAX_FS_OPT_LEN - ((entry->type == T_DIR) ? 6 : 0), 8, false);
```
d0k3: You're changing this because, otherwise, there would be leftovers from the line above / below when scrolling.

d0k3:
Okay, I had a look and I like the scrolling thing, however, some changes are problematic.

The `ShowSelectPrompt` function is meant to be universal. With your changes, if more than 8 (10) entries are in the menu, we got scrolling. Scrolling doesn't work properly cause not all lines are of equal length and there are leftovers when scrolling. You fix this for the FileSelector, but that doesn't mean it won't cause more trouble elsewhere.

I also checked the `Scripts` and `Payloads` submenus (in HOME menu), and they don't look too good now cause the menus are aligned to the left (while all other ShowSelectPrompt based menus are centered).

I'm unsure about the position bar. It's nice, but it also makes the code complicated, and for the typical short lists of options in these submenus it is really not required (although some indicator that there is more to the list is required).

So... `ShowSelectPrompt` has to stay universal (meaning, has to work with everything), and the FileSelector still needs work. One way to go about it is to make the FileSelector a separate UI function, but that's only a suggestion and I'm open for other solutions.

arm9/source/filesys/fsutil.c
windows-server-2003: Okay.

windows-server-2003:
Okay, fixed.

windows-server-2003:
I created a new ui function(ShowFileScrollPrompt).
Also removed the scroll bar and added "[%d more]" below the last entry instead.
"menus are aligned" means the message "HOME menu..." and "<A> select..." are aligned, right ?
If it does, I fixed it.

d0k3:
I'll have a look and get back later, alright?

d0k3:
Okay, first of all, sorry, this took me as long as it did.

There is still something wrong with this. Just take a look in the HOME menu -> `Scripts` or HOME menu -> `Payloads`. The stuff in there should be centered, with this pull request, stuff is aligned to the left. This is caused by the replacement of `TruncateString` with `ResizeString`.

I of course know why you're doing this (so that there are never leftovers of the names atop / below when scrolling), but you will need to find a different solution for this.

windows-server-2003:
Sorry for keeping you waiting.
Implemented it as new style file selector.

d0k3:
Had a look, and I like it. There's still room for improvement:
https://i.imgur.com/9LGAynf.png

Even with the truncation, something is cut off. You need to truncate to a shorter width. You should also test with an 8px font. Use the `FONT_WIDTH_EXT` macro to calculate the truncated length based on the font width.

I've still got two suggestions... first, with the interface being very explorer like, it's somewhat weird to see a prompt when hitting an empty folder - maybe you want the dotdot (..) entry in there instead?

The last thing really is only a suggestion and comes down to personal tastes. I think the scroll bar on the right side is not required, cause you always know how many files are still there by the bottom entry. Leave it in, if you like it (the new file selector is yours, and I'm okay with it), or reconsider.

d0k3:
Taken over locally, thanks!

PR #428, opened 2018-09-13
Improved file loading time

windows-server-2003:
The present sort algorithm for file sorting is "selection sort".
However, it's a bit slow.

Since qsort is there, you should use qsort for faster sorting.
We can't sort an array of DirEntry directly as qsort uses memcpy for swapping(we have to use DirEntryCpy to swap them) but we can sort an array of pointer of DirEntry.

It's being much faster to load many files in single directory(5s->0.5s for me).

### DIFF
--- a/arm9/source/filesys/fsdir.c
+++ b/arm9/source/filesys/fsdir.c
@@ -6,23 +6,24 @@ void DirEntryCpy(DirEntry* dest, const DirEntry* orig) {
 }
 
 int compDirEntry(const void* e1, const void* e2) {
-    const DirEntry* entry1 = *((const DirEntry**) e1);
-    const DirEntry* entry2 = *((const DirEntry**) e2);
+    const DirEntry* entry1 = (const DirEntry*) e1;
+    const DirEntry* entry2 = (const DirEntry*) e2;
     if (entry1->type == T_DOTDOT) return -1;
     if (entry2->type == T_DOTDOT) return 1;
     if (entry1->type != entry2->type)
         return entry1->type - entry2->type;
-    return strncasecmp(entry1->name, entry2->name, 256);
+    return strncasecmp(entry1->path, entry2->path, 256);
 }
 
 void SortDirStruct(DirStruct* contents) {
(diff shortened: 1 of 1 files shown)


d0k3:
Alright. While this a pretty great idea, some stuff just ain't right yet. You take too much memory from the stack there, and doing so is not even required.

Just have a look at what `DirEntryCpy` does - it just does a `memcpy` and some pretty simple fix afterwards. You could also just qsort the whole array (not pointers, actual `DirEntry`), then apply the same fix to the whole list afterwards. This way, you won't need two additional (big!) arrays on the stack,

Maybe I'm missing something here, tho, so you you tell me.

windows-server-2003:
> apply the same simple fix to the whole list afterwards
I think there's no good way to do that.
```
int org_index =  (entry->name-contents)/sizeof(DirEntry);
int cur_index = (entry - contents)/sizeof(DirEntry);
entry->name += sizeof(DirEntry)*(cur_index - org_index);
```
Here is a way to do that but I don't think it's well readable.

I have another reason I wouldn't like to do that. Qsort performs (N*log_2(N))(N*11 if N = 2048) times of memcpy and memcpying DirEntries would take time than DirEntry*s.

My solution is using malloc to allocate memory.

d0k3:
Well, actually you need to search for the `/` again to manually apply the fix to the whole list. `name` is just a pointer to the position in `path` where the actual filename starts (`strrchr` is your friend). Not that super simple, granted, and there are ways to mess up, but also not too bad.

Now... memory limitations in GM9 (or, on the 3DS console) are somewhat tight. An additional requirement of half an MiB may very well matter. Sooo... will `qsort` on actual `DirEntry` really take that much longer? Got a quick way to test? If so, using `malloc` is very much acceptable.

windows-server-2003:
It's like 0.3s->0.5s(caused by changing name comparation from ```entry->name``` to ```entry->path```). Good.

d0k3:
Taken over locally. Thanks!

urherenow:
entry undeclared. breaks build.

PR #463, opened 2019-03-14
Redo alphabet to be better

eiiiiix:
Currently renaming a file has all caps and lower grouped together and most files either use all lowercase or have only afew caps
changing the order makes it alittle easier to quickly rename for people who have bad L and R buttons

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -53,6 +53,10 @@ ifeq ($(SHOW_FREE),1)
     CFLAGS += -DSHOW_FREE
 endif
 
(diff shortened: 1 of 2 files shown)


d0k3:
Well... I'd actually suggest we use the new alphabet order in all cases, thus we don't need the NO_LR compile parameter. I like it. I think the numbers should be kept in order, tho.

eiiiiix:
i only changed the number order because starting at 9 if you go backwards is a little weird

d0k3:
Taken over locally, thanks!

PR #464, opened 2019-03-14
revert seqmemcpy and memcpy reimplementation

luigoalma:
The memcpy replacement
If need to fix up anything first, like the comments on sha.c or rsa.c, say and I'll fix up


d0k3:
Taken over locally, thanks!

PR #471, opened 2019-03-31
Update cleanup script to add *miner method files

eip618:
Hi there!
This PR adds some files which are used to setup the latest *miner methods.
It also changes the handling of the cias folder, because users are often confused as to what to do with a cia when the cias folder suddenly disappears. So this change causes the contents of cias to be deleted, but not the actual folder itself.


eip618:
i'll close this as i have just realised you addressed a similar pr from valid

d0k3:
Yup. Just to be clear, I want that pull request. But I want @annson24 to review it.

eip618:
it looks like they have noticed it and will get a handle on it shortly.

PR #474, opened 2019-04-03
Very minor optimization

luigoalma:
It only effect happens at common misaligned memory check and fixup
Does the exact same end result, just in less 2 instructions and independent of r4 and r5, using r12 instead


d0k3:
Thanks, @luigoalma, merging it now...

PR #476, opened 2019-04-08
ARM11 improvements

Wolfvak:
I'm currently working on a branch that will bring some improvements to how GM9 utilizes the ARM11 processor. It's currently a work in progress that shouldn't be merged until heavy testing is done on it (wink wink, to anyone reading this...), and until all the goals are achieved

Immediate goals:
- [x] Improve the PXI protocol to make it as reliable as possible, as well as lower the latency
- [x] Move I2C code to the ARM11
- [x] Add register caching for the MCU (reads are performed from the cache and it's only updated whenever an MCU interrupt happens)
- [x] Add a shared memory region (HID updates, Unix 64 bit timestamp to reduce RTC reads, anything else that potentially matters)
- [x] Move existing code that performs background tasks to the ARM11 (backlight checks, etc)

Later goals:
- [x] Enable the MMU, caches and other fancy MP11 features like branch prediction, return stack and instruction folding.


Wolfvak:
Okay, I believe I achieved most of what I set out to do as it currently stands.

As always, lots of testing is appreciated!
Particular points to test:
 - GM9 boots at all (kind of important if you ask me)
 - GM9 boots from all entrypoints and behaves accordingly (bootloader mode, ntrboot complains about missing aeskeydb if it's not built in, etc)
 - Being able to boot all kinds of other FIRMs (as long as it makes sense to ofc, it's senseless to boot a N3DS stock FIRM on an O3DS and such)
 - Volume slider correctly modifies the LCD brightness
 - Make sure that the keypress checks behave the same as they used to


And anything else you can think of!
If you find something is broken, please comment about it here.
If it's not about a bug, or not pertinent to this PR I'll probably remove the comment.

PR #495, opened 2019-06-26
Filename changes.

pthfdr-42:
1. LUMA uses `1:/rw/luma/` for its config files, so you should do that as well. Also, it is very messy to directly put files on the root directory.
2. Change some file extensions, so Windows software that depends on extensions know that they are FAT images and let you mount/check them.

(Also, what is the purpose of FIRM2 to FIRM7?)


ihaveamac:
firm2 to firm7 can exist if they are added to the NCSD header. I don't think they should be removed.

Wolfvak:
"ctrnand_fat.fat16.img"? That sounds a bit redundant

d0k3:
Sorry, but there's no way we're taking this over. By changing all those filenames you break an untold number of tutorials for basically no gain.

I'm okay with using the 1:/rw directory *as an alternative* though. You may do a new pull request without all those filename changes.

PR #502, opened 2019-06-30
Change support file pathes.

pthfdr-42:
It is very messy to directly put files on the root directory, so I changed the support file pathes to `0:/3ds/gm9/` (since `0:/3ds/` is used by most homebrews) and `1:/rw/gm9/` (since LUMA uses `1:/rw/luma/`).
Also I added the GM9 pathes as locations for `boot.firm` so you do not have to put them at the root directory (should only work when GM9 is installed to FIRM0, so it won't affect noobs).

### DIFF
--- a/arm9/source/filesys/support.c
+++ b/arm9/source/filesys/support.c
@@ -3,8 +3,8 @@
 #include "vram0.h"
 #include "vff.h"
 
-#define SUPPORT_FILE_PATHS  "0:/gm9/support", "1:/gm9/support" // we also check the VRAM TAR first
-#define SUPPORT_DIR_PATHS   "V:", "0:/gm9", "1:/gm9"
+#define SUPPORT_FILE_PATHS  "0:/3ds/gm9/support", "1:/rw/gm9/support" // we also check the VRAM TAR first
+#define SUPPORT_DIR_PATHS   "V:", "0:/3ds/gm9", "1:/rw/gm9"
 
 
 bool CheckSupportFile(const char* fname)
--- a/arm9/source/filesys/support.c
+++ b/arm9/source/filesys/support.c
@@ -3,8 +3,8 @@
 #include "vram0.h"
 #include "vff.h"
 
-#define SUPPORT_FILE_PATHS  "0:/3ds/gm9/support", "1:/rw/gm9/support" // we also check the VRAM TAR first
-#define SUPPORT_DIR_PATHS   "V:", "0:/3ds/gm9", "1:/rw/gm9"
+#define SUPPORT_FILE_PATHS  "0:/gm9/support", "1:/rw/gm9/support" // we also check the VRAM TAR first
+#define SUPPORT_DIR_PATHS   "V:", "0:/gm9", "1:/rw/gm9"
 
 
 bool CheckSupportFile(const char* fname)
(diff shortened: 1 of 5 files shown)


TurdPooCharger:
The opinions and points made here are my alone; this is not on behalf of the GodMode9 team.

I am a frequent GodMode9 user, passerby who read this pull request, and someone who disagrees with the idea about changing the placement of the **gm9** folder.

***

If you place the **gm9** folder within **sdmc:/3ds**, Homebrew Launcher will see it as an empty folder when there isn't an 3DSX app named as `gm9.3dsx` within it. This would add clutter to its listings.

Relocating the work folder there also adds an additional directory level you have to go into to look for it versus crowding with the other 3DSX apps and folders. There are those who have lots of homebrew titles.

Many gm9 scripts would simply break when referencing between `$[GM9OUT]` vs **0:/gm9/out**/etc.. Various DIY guides would also be rendered outdated when people become confused that **gm9** is no longer visible or not found on root. 

Luma3DS when booted from SD card has the work directory at **sdmc:/luma**. It would be weird to have **luma** folder at card root yet **gm9** is put in **smdc:/3ds**.  Some end users might see GodMode9 not sharing the same tier of prestige as Luma3DS if its own folder was relocated.

CTRNAND Luma3DS needs **1:/rw/luma** in order to hold `config.bin` when booting HOME Menu without SD card, recording error logs, and carrying FIRM payloads.

I'm having a hard time determining what sort of files CTRNAND GodMode9 may need at **1:/rw/gm9** that involves operations without the SD card.

Wolfvak:
To be fair, we do plan to have a very simple configuration file to store some persistent data (brightness, etc) and I personally wouldn't mind changing the CTRNAND path to be inside the `rw` directory, in the case where GM9 is booted without an SD (be it from FIRM0, ntrboot or from a FIRM stored in a filesystem).

Changing the SD path is a no-go for me though, there's just no point to it and I believe it'd break too much existing documentation.

I'll leave the PR open for further discussion, but don't expect too much positive feedback.

TurdPooCharger:
I am unsubscribing from this pull request as @pthfdr-42 back-to-back commit pushes are bombarding my e-mail notifications.

pthfdr-42:
@Wolfvak Fine...
However, I really dislike putting folders directly at the root of FS. Android is infamous for this, and the same can be said for all Linux programs that ignore `XDG_CONFIG_DIR`.

Wolfvak:
For the record, I only said I "[...] personally wouldn't mind changing [...]", not that I would absolutely merge when the changes were implemented.

That decision still falls on @d0k3, which is why I kept the PR open after expressing my position on this matter.

And yeah, as @TurdPooCharger said, you _really_ need to merge all future changes into a single commit with all the changes, instead of a single commit for every file. You can't change the past (well with git you can...) but try to make sure you follow this in the future.

d0k3:
Anything that breaks stuff that worked earlier won't be merged. We also don't need additional bootfirm paths. These are only relevant if GM9 is an integral part of your system (the bootloader), and the `boot.firm` file is important enough to be in the root of both, SD and CTRNAND. Even b9s handles stuff that way.

I'm okay with the `1:/rw/gm9` path *as alternative* to the `1:/gm9` path for support files.

pthfdr-42:
I added additional boot pathes so you do not *have to* put files at the root of the FS. It won't affect existing users.

d0k3:
Well, you're breaking support for `1:/gm9`. Can't megre that, as some tutorials may refer to it.

d0k3:
@pthfdr-42 - add back the support for 1:/gm9. Also do some boot speed testing (is there any noticable difference from your changes?). After that get back, and I'll merge the pr (if it doesn't destroy boot performance).

pthfdr-42:
I officially give up.

Issue #40, opened 2017-04-06
Feature Request - Save Management for DS/3DS Cartridges


JoshuaDoes:
Hello again!

I have a feature request and I'm unsure of where else to place this. I've noticed that GodMode9 has full support for dumping the ROM file in either full or trimmed format for DS cartridges instead of only 3DS cartridges, but one thing I'd like to have implemented is the ability to dump and replace the save data on the cartridge. A good place to look at for reference is [TuxSH's TWLSaveTool](https://github.com/TuxSH/TWLSaveTool) homebrew, where the feature is fully implemented as its own app in userland and also installable as an app for the system.

Thank you for all of your hard work once again!
If this cannot be done, all I ask is for an explanation as to why. However if it is just something you don't wish to focus on right now, that's also fully understandable and I wish you good luck with all else!

d0k3:
That feature is, in fact, on my own wishlist:
https://pastebin.com/Sp2GmWda

Problem is, userland code works differently and most likely won't do the trick. I could use help from someone else willing to figure this out, for 3DS game saves, too.

JoshuaDoes:
I'm glad you're at least looking into it, and having support for 3DS game saves as well would be amazing too!

Userland code definitely is different since a lot of stuff is there to help out but with programming for the actual ARM9 none of that is there, so it's completely understandable. I wish I knew enough myself to figure this out, and if I did I'd probably have submitted a pull request with it by now. Hopefully progress may come to light eventually :)

WubTheGame:
i like this idea, and i think Godmode9 will soon be the ultimate 3DS tool. besides, it's called Godmode9, the clue's in the name.

JoshuaDoes:
It already is the ultimate 3DS tool, aside from things currently done within regular homebrew such as CIA installation and DSP firm dumping (which would be neat to have in GM9, but I don't know the specifics and thus am not sure if that's possible despite our level of access).

WubTheGame:
oh cool.

RubberBrony:
@JoshuaDoes I think DSP firm dumping as it stands using DSP1 or DSPDump relies on other titles (HOME Menu?) having a valid DSP firmware in memory. It's definitely present on the NAND as well, though, probably in the files for some system title.

NamedKitten:
I would LOVE to see this implemented.

SirNapkin1334:
[...]

PR #503, opened 2019-07-04
Fix crashes when opening extensionless files.

dratini0:
I have noticed that trying to open extensionless files crashes, and that this happened after @Wolfvak's MPU fix. Adding some nullchecks seems to fix the issue.

Here is a crash report from the GM9-TD hourly:
```
Exception: Data Abort (4)
GodMode9 v1.8.0-69-g3cbbd538
2019-07-05 01:21:46
 
R00: 00000000 | R01: 08064646
R02: 00000004 | R03: F19FFFFD
R04: 227ED840 | R05: 00F0A0C8
R06: 08064646 | R07: 00000000
R08: 0806464A | R09: 00000200
R10: 227ED940 | R11: 2000135C
R12: 00000005 | R13: 227ED6B8
R14: 08034EDC | R15: 0804D1E4
CPSR: 200000DF

Stack:
227ED6B0: 40 D8 7E 22 64 18 03 08 40 D8 7E 22 C8 A0 F0 00 
227ED6C0: 65 13 00 20 5C 13 00 20 00 00 00 00 DC 4E 03 08 
227ED6D0: 00 00 00 00 53 4D 44 48 33 44 53 58 2C 00 0A 0E 

Code:
0804D1A0: 12800001 131308FF 12800001 E12FFF1E 
0804D1B0: E3520000 0A000020 E92D41F0 E1A07000 
0804D1C0: E1A06001 E0818002 EA000005 E0540005 
0804D1D0: 1A000014 E3550000 0A000014 E1580006 
0804D1E0: 0A000012 E4D74001 EBFFFB93 E0800004 
0804D1F0: E5D03001 E2033003 E3530001 02844020 
0804D200: E4D65001 EBFFFB8C E0800005 E5D03001 
0804D210: E2033003 E3530001 1AFFFFEB E2850020 
```

### DIFF
--- a/arm9/source/filesys/filetype.c
+++ b/arm9/source/filesys/filetype.c
@@ -132,7 +132,7 @@ u64 IdentifyFileType(const char* path) {
     } else if ((strncasecmp(ext, "png", 4) == 0) &&
         (fsize > sizeof(png_magic)) && (memcmp(data, png_magic, sizeof(png_magic)) == 0)) {
         return GFX_PNG;
-    } else if (ext && ((strncasecmp(ext, "cdn", 4) == 0) || (strncasecmp(ext, "nus", 4) == 0))) {
+    } else if ((strncasecmp(ext, "cdn", 4) == 0) || (strncasecmp(ext, "nus", 4) == 0)) {
         char path_cetk[256];
         char* ext_cetk = path_cetk + (ext - path);
         strncpy(path_cetk, path, 256);
@@ -150,7 +150,7 @@ u64 IdentifyFileType(const char* path) {
         return BIN_LEGKEY; // legacy key file
     } else if (ValidateText((char*) data, (fsize > 0x200) ? 0x200 : fsize)) {
         u64 type = 0;
-        if ((fsize <= SCRIPT_MAX_SIZE) && ext && (strncasecmp(ext, SCRIPT_EXT, strnlen(SCRIPT_EXT, 16) + 1) == 0))
+        if ((fsize <= SCRIPT_MAX_SIZE) && (strncasecmp(ext, SCRIPT_EXT, strnlen(SCRIPT_EXT, 16) + 1) == 0))
             type |= TXT_SCRIPT; // should be a script (which is also generic text)
         if (fsize < STD_BUFFER_SIZE) type |= TXT_GENERIC;
         return type;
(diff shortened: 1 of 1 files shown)


Wolfvak:
This looks correct, but would you mind simply making the function return unknown whenever there is no extension present? Just to make it more future proof.

Also, thanks for noticing, one of the points of the rewritten MPU and exception handling code was to prevent NULL reads.

aspargas2:
The majority of checks in that function don't rely on file extension or filename at all, so I personally don't see any reason there shouldn't be detection for extension-less files in all the file types that it's able to recognize without an extension present.

dratini0:
I agree, that would create an odd corner case where (for example) fonts are recognized when they have the wrong extension, but not when they have none. Not that either case should happen often.

aspargas2:
There are also some cases where a supported file would normally not have an extension, such as extdata files (recognized as DIFF) and `cetk`s and `tmd`s from the CDN.

Wolfvak:
What about making it point to an empty ("") string instead? That would still work, right?
My problem is if a new type is implemented then it will have to check against NULL, too, but they might forget to add the check and the bug resurfaces.

aspargas2:
Yeah I guess that would work, just adding the following after line 27:
```C
if (!ext) ext = "";
```

dratini0:
That makes sense. Nulls are a terrible thing.

aspargas2:
Doesn't that change make the first commit unnecessary?

Wolfvak:
Remember there's no need to check for the pointer being NULL, so would you mind removing those 2 checks you added earlier? Also, try to move the fix after the last variable declaration, with proper indentation (looks weird and could confuse people otherwise)

dratini0:
Sure, but, by the same token, wouldn't I remove all the previous ones as well? There were 2 or 3 already.

dratini0:
I did these things, you are free to cherry-pick as you like (if just to flatten the commit, even...). I will be busy messing with SPI drivers.

profi200:
Regarding SPI it has been fully figured out: https://github.com/derrekr/fastboot3DS/blob/master/source/arm9/hardware/spicard.c

That doesn't implement quad SPI though since the speed gain will be low and we don't know the cmds for each flash type to enable this mode. What's missing is the actual SPI flash code. Also mentioning #40

dratini0:
Oooh, that is neat. I will give it a read in a moment. (I am trying to fix cartsaves, which seemed to have an issue with the SPI driver. See #40 )

dratini0:
Neat, thank you! I will be looking at it in the next few hours!

Wolfvak:
Sorry it took so long to merge, I kinda missed the new commits.

Thanks for the fix, and make sure to let us know if you find anything else!

dratini0:
No problem!

PR #504, opened 2019-07-06
Get rid of extra copy of readme

dratini0:
See 9ab9c01


Wolfvak:
Well that was a dumb bug. Thanks!

PR #509, opened 2019-07-15
DISA/DIFF writing and title database inner FAT support

aspargas2:
This adds support for writing to DISA and DIFF containers, and for manipulating the inner FAT filesystem of title database files (this includes `ticket.db`). This does not directly add anything usable to the end user, just the code to make such things. I imagine this could eventually be turned into things like having an `Install as title` option on CCI, NCCH, and CIA files, and support for mounting title database files with full write access. This code could also be modified to work with the inner FAT filesystems of savegames and extdata, which I may do myself at some point.

If master is not the appropriate branch to which to be making the PR, given its nature of not adding any immediately useful functionality, just let me know what would be appropriate.

Also, credits go entirely to @wwylele for reverse engineering and documenting the title database stuff on [3DBrew](https://www.3dbrew.org/wiki/Inner_FAT).


Wolfvak:
What are potential implementation pitfalls? For example, I remember if you didn't have any title installed then FBI would fail to install any CIAs until you installed something from the eShop - something about internal databases not existing yet.

Also, how well has this been tested? The code looks great from a quick glance, but could it misbehave on something exotic like developer consoles?

Finally, could you add a sort of "BDRI Context" feature? Just for performance reasons, that way you don't have to keep opening and closing files, which incur in costly FAT and flash operations. A possibility would be to have that context keep a pointer to the open file handle, plus any other needed data and functions to create the context (open the file, retrieve any data), manipulate it (ReadTitleInfoEntryFromDB, ...) and finally flush/destroy it. Just an idea.

ihaveamac:
>if you didn't have any title installed then FBI would fail to install any CIAs until you installed something from the eShop

FBI calls [AM_InitializeExternalTitleDatabase(false)](https://smealum.github.io/ctrulib/am_8h.html#a469ba855d3817fa86091baa17d136676) to create the database if it doesn't exist.

Until there's a way to generate the title database, one lazy method would be for something to include an empty title.db for use. 3DS Hacks Guide could do this.

aspargas2:
Yeah, this code is currently unable to create a DISA/DIFF file from scratch or initialize the inner FAT from scratch, so any feature to install a title (such as the "CartInstall" fork I made using this code) would currently have the pitfall you described. One way to fix that would be to just embed a blank title.db that gets copied into the proper location if one doesn't already exist, but that's not really a proper solution. A proper solution would involve being able to create an artificial DIFF file of an arbitrary internal size and initialize a BDRI filesystem with arbitrary values for pretty much everything in the FS header, which is all doable, but would require a non-negligible amount of additional development and testing.

This code has only been tested on retail consoles, but it has been tested on them enough times that I'm reasonably confident there aren't any major bugs in its functionality on them. The only part of it that has not been tested is writing to DISA files; when designing that code I only had DIFF in mind, but later realized there was no reason it wouldn't also work with DISA. As for developer systems, I have almost no experience or knowledge on what's different about them and their database files, so I have no idea if this code would work on them, as I don't know if there's something fundamentally different about their version of DIFF and BDRI.

I agree opening and closing the file every time an entry read, removed, or added isn't optimal, but I think a better overall solution would be to add support to mount arbitrary DISA/DIFF files, resulting in a virtual file to represent the contents of IVFC lvl 4, then change the BDRI code to operate on a file using fvx_ calls rather than DisaDiff ones.

aspargas2:
Also, as I already mentioned in IRC but forgot to mention here, this code isn't able to work with a couple special case tickets which are not the usual 0x350 bytes, so the `Ticket` struct in GodMode9 wouldn't apply to them. The functions in my code that use that struct are effectively just wrappers for ones that can handle entries of arbitrary sizes, though, so modifying my code to handle such special case tickets would be trivial.

wwylele:
Glad to see the work here is being done. Sadly I don't have time to help code reviews in detail. As one who has implemented DISA/DIFF/InnerFAT stuff three times, I want to give some high-level advise:
 - Write unit tests early, as this stuff is too complicated to get right on the first try. You want to find mistakes early before screwing up user's 3DS data. I wrote my unit test as a program that randomly reads/writes/adds/removes items in the file system and verify the content.
 - Think optimization early. The DPFS+IVFC(+upper layer AES encryption) data format can cause really bad performance if the program is written in a naive way. For example, avoid fixing the IVFC hash chain on every write operation, because adjacent write operations will result in fixing the same hash multiple times on the upper levels. Try clustering them to, for example, when closing the file.

My latest attempt is here if you didn't know yet: https://github.com/wwylele/save3ds/tree/master/libsave3ds/src (Sorry for coding in rust). I hope it can help.

aspargas2:
I appreciate the advice, and all your work in reverse engineering the files initially and writing a working tool to modify them. Your tools were very helpful in testing this code.

To the first point: I definitely agree with that there would be no way to get this right on the first try. There was a long cycle of trial and error and silly mistakes with both the DIFF and the inner FAT side of things. This code being in only two commits is the result of a lot of squashing and a lot of not committing things till they worked.

To the second point: I admittedly did not have optimization as a huge priority when writing this. I did exactly the thing you just advised not to do, just because at that time I didn't foresee this code being used for anything more than a few writes at a time, and @Wolfvak already pointed out another way in which it could be better optimized. I intend to implement both of those suggestions now that this code is looking at going into a context where it could be used for any number of things of various intensity, so I'm going to ask that this not be merged until I do that.

And yeah, I know of save3ds; its existence was the inspiration for me to attempt to write this in the first place. I don't know any rust, though, which I'm actually kind of glad about because it forced me to write this just from 3DBrew rather than sloppily porting half of another tool's code to work for this context.

d0k3:
@aspargas2 - you get back once you deem it ready to merge, alright? I'll miss it otherwise. I definitely want this in the next GM9 release.

aspargas2:
@d0k3 Will do. I've just had a lot of IRL stuff going on this week and the last, but I should have a lot more free time next week to work on this.

d0k3:
Any news @aspargas2 ? This is still not ready to merge, right?

aspargas2:
@d0k3 Yeah, it's still not ready. I'll let you know when it's ready for another review. Really sorry about all the delays; my 3DS is not in a working state right now and with school starting I haven't had a lot of time to work on this.

aspargas2:
@d0k3 @Wolfvak I think this is ready for another review. It currently only caches 32 writes, but changing that is just changing the number on line 9 of `vdisadiff.c`. Your wish is my command on what number you want to put there.

aspargas2:
Ok, should be ready now (third time's the charm, right?)

d0k3:
Alright, merged to master. Thanks a lot, that's real good stuff! Looking forward to additional work you may be doing on this.

PR #519, opened 2019-08-05
added essential.exefs dump to the boot9&11 script

GlaZedBelmont:


d0k3:
That's not so good, actually. If essential.exefs is not present, this prevents you from dumping the bootroms. Maybe dumping essential.exefs via another menu entry would be better? I'd also like to read @annson24 's take on this.

d0k3:
@GlaZedBelmont - I can't take that pull request without the changes I requested. Want to still do them?

GlaZedBelmont:
yeah, hold up, ill do the changes real quick.

something like if it doesnt find essential.exefs, it asks you if you still want to dump the bootroms

d0k3:
Okay, sorry to have kept you waiting that long!

It would still be better to have a separate entry for dumping essential.exefs. Can you do that?

Normal course of action would be to first do a pull request on @annson24's repo.
https://github.com/annson24/GM9Megascript
Dunno if @annson24 (pingpingping) is still active, so it's okay to do pull requests here, too.

d0k3:
Okay, I had a look over your changes and also tested the script. Your changes introduced some display errors ( strings too long) and, well, it should have been a separate entry all along. I tried to separate your changes into a new menu entry myself, but failed.

As we're nearing a new release and I actually want that feature in, I went ahead and added the menu entry and code myself. I still want to thank you for your contribution, we just couldn't merge it this time.

GlaZedBelmont:
No problem, I PRed the script itself instead of PRing GM9

GlaZedBelmont:
also, if you could add me and win9x to the script's credits, we both worked on the idea and creation of the script :smiley:

PR #545, opened 2019-10-12
Update user manual

DMSalesman:


Wolfvak:
honestly, wouldn't it be easier to just merge the script used to do this instead of having to ask you each time there's a release with an updated readme?

DMSalesman:
Sure, but I'm in no way proficient enough with `make` to risk submitting a pull request, at the moment.

Wolfvak:
if you use a python script just merge that into the utils dir and I'll add it to the makefile

d0k3:
Merged, thank you!

DMSalesman:
@Wolfvak The Python script is what does the job, but there's also a "patch database" consisting of a `gzip`ped .json, which takes care of the more advanced search and replaces. Is there a way to handle that?

Wolfvak:
I suppose it could be placed in the same dir as the script?

DMSalesman:
@Wolfvak #546.

PR #546, opened 2019-10-12
Add Unmark as an external tool

DMSalesman:
`unmark` is a tool to produce an user manual from the `README.md`.

The `patch.json.gz` file contains a list of patterns to match and
their replacements.
If `unmark` complains about being unable to match a pattern, please open
an issue with the warning and your `README.md` at
[upstream](https://github.com/DMSalesman/Unmark).


Wolfvak:
looks sane, will merge and add to the makefile to regenerate the readme on release target

PR #547, opened 2019-10-12
added credits from the essential.exefs script

GlaZedBelmont:
we both thought and wrote the script, so I thought the addition of the names would be great


DMSalesman:
Unrelated, but it seems "8bitwonder" is mentioned twice.

Wolfvak:
Merged. Also, thanks for noticing that @DMSalesman, will fix up in the next commit.

Issue #494, opened 2019-06-26
Handling GBA VC .sav CMAC


TurdPooCharger:
Continuing what was left off from,

- [[Research needed] Handling .cmd CMACs on SD card](https://github.com/d0k3/GodMode9/issues/340)

this should solve transferring the entirety of the **Nintendo 3DS** folder, minus games that have saves with Secure Value (and saves made outside that folder).

Edit - Forgot to add the 'ARM7 registers' field.
![GBA VC diagram](https://user-images.githubusercontent.com/35666439/60226946-64acc180-985b-11e9-9098-d523810319c6.png)

![GBA VC cmac calculation](https://user-images.githubusercontent.com/35666439/60145043-c5bc9280-9792-11e9-8914-4873f86d2219.png)

***

I have more findings to share later when I get the chance to. Check back if you're interested in direct extraction and injection of GBA saves without the involvement of `agbsave.bin`.

TurdPooCharger:
@BernardoGiordano / FlagBrew team / Checkpoint
[[Checkpoint] GBA game save support. #22](https://github.com/FlagBrew/Checkpoint/issues/22)

@J-D-K (JK) / JKSM

***

The `.gba` rom for a GBA Virtual Console title can be found at:

- **[A:] SYSNAND SD/title/00040000/<TID_LOW>/content/**
- `0000000#.app` → **NCCH image options...** → **Mount image to drive**
- **G:/exefs** → `.code`

The `.code` consists of the gba rom with a `.CAA` footer. This footer is 864 (0x360) bytes.

![CAA footer](https://user-images.githubusercontent.com/35666439/60205111-5e4c2480-981e-11e9-8bef-d6754e632844.png)

***

The Game Boy Advance has fifteen (15) different save types.

- `0x0`: EEPROM 8k
- `0x1`: EEPROM 8k (?)
- `0x2`: EEPROM 64k
- `0x3`: EEPROM 64k (?)
- `0x4`: Flash 512k (Atmel, ID: 0x3D1F) + RTC
- `0x5`: Flash 512k (Atmel, ID: 0x3D1F)
- `0x6`: Flash 512k (SST, ID: 0xD4BF) + RTC
- `0x7`: Flash 512k (SST, ID: 0xD4BF)
- `0x8`: Flash 512k (Panasonic, ID: 0x1B32) + RTC
- `0x9`: Flash 512k (Panasonic, ID: 0x1B32)
[...]

PR #548, opened 2019-10-17
Add handling for SD GBA VC .sav CMAC. (Fixes issue #494).

TurdPooCharger:
When identifying a selected `00000001.sav`, cmac type testing is first attempted for 3DS saves with the **DISA** format. If a 3DS save data is not detected, the search continues at expected offsets for gba vc bottom slots in increasingly larger save sizes. For the GBA VC saves, this only calculates and corrects whichever of the two slots is newer or has the higher counter. There are five additional **CMAC_AGBSAVE_** types to differentiate the different gba save sizes. If a bottom slot was found but determined to be older based on comparing the counter values, the search defaults to **CMAC_AGBSAVE_SD**.

For the save counters, there are two scenarios where the value `00` is considered bigger than `FF` depending which slot has what. If clarification is required, I will post a diagram at [issue #494](https://github.com/d0k3/GodMode9/issues/494) explaining what is meant by `00` bigger than `FF`.

Please make corrections and improvements to the coding as you see fit in your peer review. I am not well versed or efficient in the C/C++ programming language.

### DIFF
--- a/arm9/source/utils/nandcmac.c
+++ b/arm9/source/utils/nandcmac.c
@@ -66,34 +66,31 @@ u32 SetupSlot0x30(char drv) {
     
     return 0;
 }
-
 u32 FindAgbSaveSlotOffset(const char* path, u32 cmac_type) {
-    u8 upper_counter[1];
-    u8 lower_counter[1];
+    u32 upper_counter[1];
+    u32 lower_counter[1];
     u32 magic_check[1];
-    u32 slot_offset = 0;
-    u32 bottom_offset[5] = {0x400, 0x2200, 0x8200, 0x10200, 0x20200};
+    u32 slot_offset[6] = {0, 0x400, 0x2200, 0x8200, 0x10200, 0x20200};
     u32 i;
 
-    if (cmac_type == CMAC_AGBSAVE) return (slot_offset); // Does not apply for 'agbsave.bin'.
+    if (cmac_type == CMAC_AGBSAVE) return slot_offset[0]; // Does not apply for 'agbsave.bin'.
     
-    for (i = 0; i <= 4; i++) { // Look for the `.SAV` magic header at the expected bottom slots.
-        if (fvx_qread(path, magic_check, bottom_offset[i], 0x4, NULL) != FR_OK) return 0;
+    for (i = 1; i <= 5; i++) { // Look for the `.SAV` magic header at the expected bottom slots.
+        if (fvx_qread(path, magic_check, slot_offset[i], 0x4, NULL) != FR_OK) return 0;
         if (*magic_check == 0x5641532E) break; // Magic header '.SAV' found.
     }
 
-    if (i == 5) return 0; // Bottom slot not found.
+    if (i == 6) return 0; // Bottom slot not found.
 	
     // Compare top and bottom slots' counter values to determine which is newer.
-    if (fvx_qread(path, upper_counter, 0x34, 0x1, NULL) != FR_OK) return 0;
-    if (fvx_qread(path, lower_counter, bottom_offset[i]+0x034, 0x1, NULL) != FR_OK) return 0;
+    if (fvx_qread(path, upper_counter, 0x34, 0x4, NULL) != FR_OK) return 0;
+    if (fvx_qread(path, lower_counter, slot_offset[i]+0x034, 0x4, NULL) != FR_OK) return 0;
 
-    if ((*upper_counter == 0xFF) && (*lower_counter == 0x00)) slot_offset = bottom_offset[i]; // Scenario #1: First save is initialized. Bottom slot is newer.
-    else if ((*upper_counter == 0x00) && (*lower_counter == 0xFF)) slot_offset = 0;           // Scenario #2: (ie, 0x100 > 0xFF). Top slot is newer.
-    else if (*upper_counter > *lower_counter) slot_offset = 0;                                // Scenario #3: Top slot is newer.
-    else slot_offset = bottom_offset[i];                                                      // Scenario #4: Bottom slot is newer -or- both are the same. 
+    if (*upper_counter == 0xFFFFFFFF);                // Scenario #1: First save is initialized. Bottom slot is newer.
+    else if (*upper_counter > *lower_counter) i = 0;  // Scenario #2: Top slot is newer.
+    // else;                                          // Scenario #3: Bottom slot is newer -or- both are the same. 
 
-    return (slot_offset);
+    return slot_offset[i];
 }
 
 u32 CheckAgbSaveHeader(const char* path) {
--- a/arm9/source/utils/nandcmac.c
+++ b/arm9/source/utils/nandcmac.c
@@ -11,19 +11,24 @@
 // https://3dbrew.org/wiki/Savegames#AES_CMAC_header
 // https://3dbrew.org/wiki/Nand/private/movable.sed
 // https://3dbrew.org/wiki/3DS_Virtual_Console#NAND_Savegame
-#define CMAC_EXTDATA_SD     1
-#define CMAC_EXTDATA_SYS    2
-#define CMAC_SAVEDATA_SYS   3
-#define CMAC_SAVE_GAMECARD  4
-#define CMAC_SAVEGAME       5 // this is not calculated into a CMAC
-#define CMAC_SAVEDATA_SD    6
-#define CMAC_TITLEDB_SYS    7
-#define CMAC_TITLEDB_SD     8
-#define CMAC_MOVABLE        9
-#define CMAC_AGBSAVE       10
-#define CMAC_AGBSAVE_SD    11
-#define CMAC_CMD_SD        12
-#define CMAC_CMD_TWLN      13 // unsupported
+#define CMAC_EXTDATA_SD           1
+#define CMAC_EXTDATA_SYS          2
+#define CMAC_SAVEDATA_SYS         3
+#define CMAC_SAVE_GAMECARD        4
+#define CMAC_SAVEGAME             5 // this is not calculated into a CMAC
+#define CMAC_SAVEDATA_SD          6
+#define CMAC_TITLEDB_SYS          7
+#define CMAC_TITLEDB_SD           8
+#define CMAC_MOVABLE              9
+#define CMAC_AGBSAVE             10
+#define CMAC_AGBSAVE_SD          11
+#define CMAC_CMD_SD              12
+#define CMAC_CMD_TWLN            13 // unsupported
+#define CMAC_AGBSAVE_EEPROM_512  14
+#define CMAC_AGBSAVE_EEPROM_64K  15
+#define CMAC_AGBSAVE_SRAM_256K   16
+#define CMAC_AGBSAVE_FLASH_512K  17
+#define CMAC_AGBSAVE_FLASH_1M    18
 
 // see: https://www.3dbrew.org/wiki/Savegames#AES_CMAC_header
 #define CMAC_SAVETYPE NULL, "CTR-EXT0", "CTR-EXT0", "CTR-SYS0", "CTR-NOR0", "CTR-SAV0", "CTR-SIGN", "CTR-9DB0", "CTR-9DB0", NULL, NULL, NULL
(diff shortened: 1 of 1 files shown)


TurdPooCharger:
@d0k3, the `nandcmac.c` I worked on does not include this [commit](https://github.com/d0k3/GodMode9/commit/f24db1442e8f46bfc29891d08eb26c6f30ba8501) you made several days ago.

My computer setup is having issues compiling the latest hourlies, so the coding and testing was done on a slightly older build where `make` outputs FIRM. I'll probably nuke my Manjaro and start fresh again, even though it and devkitarm was installed less than a month ago.

Wolfvak:
This should probably be merged into it's own branch in the GM9 repo, not `staging`

arm9/source/utils/nandcmac.c
```diff
@@ -99,7 +194,12 @@ u32 ReadWriteFileCmac(const char* path, u8* cmac, bool do_write) {
     
     if (!cmac_type) return 1;
     else if (cmac_type == CMAC_MOVABLE) offset = 0x130;
-    else if ((cmac_type == CMAC_AGBSAVE) ||  (cmac_type == CMAC_AGBSAVE_SD)) offset = 0x010;
+    else if ((cmac_type == CMAC_AGBSAVE) || (cmac_type == CMAC_AGBSAVE_SD)) offset = 0x010;
+    else if (cmac_type == CMAC_AGBSAVE_EEPROM_512) offset = 0x410;
+    else if (cmac_type == CMAC_AGBSAVE_EEPROM_64K) offset = 0x2210;
+    else if (cmac_type == CMAC_AGBSAVE_SRAM_256K) offset = 0x8210;
+    else if (cmac_type == CMAC_AGBSAVE_FLASH_512K) offset = 0x10210;
+    else if (cmac_type == CMAC_AGBSAVE_FLASH_1M) offset = 0x20210;
```
Wolfvak: These offsets can be integrated into the information structure mentioned above.

arm9/source/utils/nandcmac.c
```diff
@@ -196,6 +302,76 @@ u32 CalculateFileCmac(const char* path, u8* cmac) {
...
+        if (!agbsave) return 1;
+        if ((fvx_qread(path, agbsave, 0x20200, AGBSAVE_MAX_SIZE, &br) != FR_OK) || (br < 0x200) ||
+            (ValidateAgbSaveHeader(agbsave) != 0) || (0x200 + agbsave->save_size > br)) {
+            free(agbsave);
+            return 1;
+        }
+        
+        u32 ret = FixAgbSaveCmac(agbsave, cmac, path);
+        free(agbsave);
+        return ret;
```
Wolfvak: Same issue here, all the code is exactly the same for each of the CMAC_AGBSAVE_X types, except for the qread offset parameter.

arm9/source/utils/nandcmac.c
```diff
@@ -67,15 +72,105 @@ u32 SetupSlot0x30(char drv) {
...
+	if (fvx_qread(path, lower_counter, 0x20234, 0x1, NULL) != FR_OK) return 1;
+    
+	if ((*upper_counter == 0xFF) && (*lower_counter == 0x00)); // First save initialized. Bottom slot is newer.
+	else if ((*upper_counter == 0x00) && (*lower_counter == 0xFF)) return 1; // Top slot is newer.
+	else if (*upper_counter > *lower_counter) return 1; // Compare top and bottom slots.
+	
+    if ((fvx_qread(path, &agbsave, 0x20200, 0x200, &br) != FR_OK) || (br != 0x200)) return 1;
+	
+	return ValidateAgbSaveHeader(&agbsave);
+}
```
Wolfvak: All the `CheckAgbSaveHeaderX` functions have too much repeated code - reading the `upper_counter` and the `if..else if block` is always the same, as is the return expression. The only things that vary are the offset for the `lower_counter` and `agbsave` read, and that information can easily be consolidated into a single structure.

There should be a single `CheckAgbSaveType` that tries to verify if it's any of the types and once it gets a match, return the type (or return a type that indicates it's not an AGB save).

d0k3:
@TurdPooCharger - don't bother too much about stuff like missing commits now. You pr needs some refactoring, though, as @Wolfvak wrote. That could be done in a lot less code. Also, this will be in the next release.

arm9/source/utils/nandcmac.c
```diff
@@ -67,15 +67,72 @@ u32 SetupSlot0x30(char drv) {
...
+        if (*magic_check == 0x5641532E) slot_offset = 0x8200;
+
+        if (fvx_qread(path, magic_check, 0x10200, 0x4, NULL) != FR_OK) return 1;
+        if (*magic_check == 0x5641532E) slot_offset = 0x10200;
+
+        if (fvx_qread(path, magic_check, 0x20200, 0x4, NULL) != FR_OK) return 1;
+        if (*magic_check == 0x5641532E) slot_offset = 0x20200;
+    }
+	
+    if ((fvx_qread(path, &agbsave, slot_offset, 0x200, &br) != FR_OK) || (br != 0x200)) return 1;
```
Wolfvak: `return 1` isn't a great idea when you're already returning other integers - try to `return 0` on error, and handle that sanely whenever `CheckAgbSaveHeader` is called.

That goes for all the `return 1`s above, too.

arm9/source/utils/nandcmac.c
```diff
@@ -67,15 +67,72 @@ u32 SetupSlot0x30(char drv) {
...
+    if ((*upper_counter == 0xFF) && (*lower_counter == 0x00)); // Scenario #1: First save is initialized. Bottom slot is newer.
+    else if ((*upper_counter == 0x00) && (*lower_counter == 0xFF)) slot_offset = 0; // Scenario #2: (ie, 0x100 > 0xFF). Top slot is newer.
+    else if (*upper_counter > *lower_counter) slot_offset = 0; // Scenario #3: Top slot is newer.
+
+    return (slot_offset);
+}
+
+u32 CheckAgbSaveHeader(const char* path) {
     AgbSaveHeader agbsave;
+    u32 magic_check[0x4];
```
Wolfvak: `magic_check` shouldn't be an array of 4 `u32`s - it's either a single `u32` or an array of 4 `u8`s (I'd say go for a single `u32` and use the `&` operator)

arm9/source/utils/nandcmac.c
```diff
@@ -67,15 +67,72 @@ u32 SetupSlot0x30(char drv) {
...
+	
+    // Compare top and bottom slots' counter values to determine which is newer.
+    if (fvx_qread(path, upper_counter, 0x34, 0x1, NULL) != FR_OK) return 1;
+    if (fvx_qread(path, lower_counter, slot_offset+0x034, 0x1, NULL) != FR_OK) return 1;
+
+    if ((*upper_counter == 0xFF) && (*lower_counter == 0x00)); // Scenario #1: First save is initialized. Bottom slot is newer.
+    else if ((*upper_counter == 0x00) && (*lower_counter == 0xFF)) slot_offset = 0; // Scenario #2: (ie, 0x100 > 0xFF). Top slot is newer.
+    else if (*upper_counter > *lower_counter) slot_offset = 0; // Scenario #3: Top slot is newer.
+
+    return (slot_offset);
```
Wolfvak: The code in this entire function is very similar to the one in `CheckAgbSaveHeader`, try making it so that one calls the other to gather relevant info.

TurdPooCharger:
About the insanity code with too many `if` and `else if`, the only programming I do these days
is GM9 scripting. The basics like `for` and `while` loops are overlooked when one is terribly
out of practice and does not code with real or more advanced languages on a consistent basis. 

***

I can't think of a good way for the `CheckAgbSaveHeader` and `FindAgbSaveSlotOffset` functions
calling from another function that shares the same list of offsets when testing or finding the
newer save slots.

For `CheckAgbSaveHeader`, the top slot at 0x0 is (usually) the first available slot that fits the
agbsave criterias for all save types. The bottom offsets are checked in case the game has made only one save where the top slot doesn't exist yet.

For `FindAgbSaveSlotOffset`, the top slot's offset is already a given. Because of that rare scenario
where the top slot hasn't been filled in yet, there is no reliable way of knowing ahead of time where
that bottom slot is located other than to search for the `.SAV` header at those possible bottom offsets.
I suppose there's a way to look for save type info in the `*.app` that has the `.CCA` footer, which is
super messy looking in the `*.tmd` to find that game `*.app`. Both of those files can be missing..

One approach is to adjust `CheckAgbSaveHeader` to only search for bottom slots to match the offset list found in `FindAgbSaveSlotOffset`, so that they can share the same list. I don't recommend this as ignoring the top slot increases wasteful computing.

Another approach is to the find bottom slot in `CheckAgbSaveHeader`, compare which slot has the higher counter, and then return or pass the offset value back to `CalculateFileCmac` and `ReadWriteFileCmac` along with the cmac_type. This basically eliminates `FindAgbSaveSlotOffset` but now has `CheckAgbSaveHeader` returning two outputs. I don't believe this can done in a clean fashion without some rewrite to `CalculateFileCmac` and `ReadWriteFileCmac` to accept the additional input.

***

For the `magic_check` variable regarding the `&` operator, I kept the `*` in place as any combination of `&magic_check` kept giving a compiling error saying mismatch between pointer and reference. 

Referring to my digital copy of "The C Programming Language, 2nd Edition by Kernighan and Ritchie",
under Section 5.1 Pointers and Addresses, page 94:

> The & operator only applies to objects in memory: variables and array elements. It
> cannot be applied to expressions, constants, or register variables.

None of the online examples I've seen has `&example_variable` used in conjunction with the (Section 2.6) equality operator: `==`.

To be honest, I never completely grasped the concepts of C/C++ pointers.

***

About the `return 1` used in `FindAgbSaveSlotOffset`, I was going off by how the `fvx_qread` used in other functions seem to be pair `return 1` for read failure(?). They're now all changed to `return 0`.

For `CheckAgbSaveHeader`, the `return 1` were kept in place as, if I'm interperting that function correctly, `return 1` states `cmac_type` is not `CMAC_AGBSAVE_SD` while `return 0` does.

Wolfvak:
Maybe we can merge it as is and fix it on our side, I'll look into it later. As long as the algorithm does what it's supposed to do it should be fine.

Wolfvak:
Ok, recent changes made the PR acceptable.
Do you have any test cases for us to try? After those are confirmed to work, we'll merge this to the `staging` branch.

TurdPooCharger:
**About test cases** (quick refresher)
- Structure of the gba vc save: [Diagram A](https://user-images.githubusercontent.com/35666439/60226946-64acc180-985b-11e9-9098-d523810319c6.png)
**Note:** The statement about the (purple) highlighted bottom slot being unimportant is incorrect.

- The five (5) save sizes: [Diagram B](https://user-images.githubusercontent.com/35666439/62171151-fe084100-b2fb-11e9-8675-04374d55ff66.png)
- How they change as the saves are made: [Diagram C](https://user-images.githubusercontent.com/35666439/61557937-a502f800-aa33-11e9-9890-34f836def380.png)

**Procedure**
1. Install a fresh example GBA VC game. Don't launch the game.
2. Go to GodMode9 and try to **Calculate CMAC**. The option should not be available.
3. Go to HOME Menu. Launch the game and make a save.
4. Go back to GodMode9 and try **Calculate CMAC**. This time, the bottom slot should be picked up.
5. Repeat step 3.
6. Repeat step 4. You should now see cmac calculation for top slot.

Issue # 494 was originally started because of interest in "SDTransfer** or moving the **Nintendo 3DS** folder between different 3DS systems without doing a System Transfer (ie, exchanging `movable.sed`).

If you have a second 3DS system, make several saves for gba vc games. Extract copies of the **dbs**, **extdata**, and **title** folders from **[A:] SYSNAND SD**. Drop them in the SD card for the target system and paste to its [A:] drive. Do **Fix CMACs for drive**. If successful, the latest saves should be picked up on the target system. 

**Manually comparing the CMACs between the two slots**
1. `00000001.sav` → **Show in Hexeditor**
2. Hold (R)-shoulder and press (Y) to change to 16 columns view.
3. Press (X) → **Search for string** → double tap **CAPS** and type in `.SAV`.
4. Hold (R)-shoulder and press (X).

d0k3:
I'll review this as quick as I can. In the meantime - @TurdPooCharger, did you also consider custom GBA VC titles with the extra large saves (forgot which games this applies for). I think in these cases, the SD save actually grows in size.

TurdPooCharger:
@d0k3, can you clarify by what you mean about gba vc titles with extra large saves? Are you referring to titles that have FLASH 1Mbit saves? The pr `nandcmac.c` does handle that size.
- [[r/3dshacks, AuroraWright] 1 Mbit AGB_FIRM lennies ( ͡° ͜ʖ ͡°)](https://old.reddit.com/r/3dshacks/comments/6znqqg/1_mbit_agb_firm_lennies_%CA%96/)

Maybe you mean games with SRAM patched saves? I have not checked those, but could find out if injection and cmac testing would be any different than games that originally have SRAM.

***
In the latest `staging` for commit c38c99e on `gameutilc`, am I checking for a difference in how the raw `*.sav` are injected to the slots found in `0000001.sav`?
- `agbsave.bin` → **Inject GBA VC save**

Is this commit to be used in conjunction with the pr `nandcmac.c`, to see if cmacs are fixed for both slots for an individual `0000001.sav`?

d0k3:
> @d0k3, can you clarify by what you mean about gba vc titles with extra large saves? Are you referring to titles that have FLASH 1Mbit saves? The pr `nandcmac.c` does handle that size.

I meant FLASH 1Mbit saves. There is no original GBA VC title (as in, available on the eShop) that has this.

> In the latest `staging` for commit [c38c99e](https://github.com/d0k3/GodMode9/commit/c38c99e5420e4f232e3b5c22d06f3475566b9fa8) on `gameutil.c`, am I checking for a difference in how the raw `*.sav` are injected to the slots found in `0000001.sav`?
> 
> * `agbsave.bin` → **Inject GBA VC save**

I'm not sure if I got that right... The GBA VC save injector needs to recalculate the CMAC cause the number (of times saved) changed between the slots, thus the CMAC changed as well.

> Is this commit to be used in conjunction with the pr `nandcmac.c`, to see if cmacs are fixed for both slots for an individual `0000001.sav`?

Also not sure if I got that right. Sure, it would be best to fix both slots. Fixing only the current one is an option, too.

d0k3:
Okay, merging this on a new staging branch, thus closing. @TurdPooCharger - you can still continue the discussion in the closed pr.

TurdPooCharger:
Not sure whether to open another pull request, so I'll post this instead.
https://github.com/TurdPooCharger/GodMode9/commit/28bf7f8947d754b660e1f0d3b95a0414c1835cd1

***
On a slightly related note about `nandutil.c` in commit [624ac57](https://github.com/d0k3/GodMode9/commit/624ac571e4815d28b34bf89563277cbd6fe56e8b), here's a proposed injection method.

When doing the **Inject GBA VC save**,

1. Record the counter value in the `agbsave.bin`.
2. Take a look in the **A:/title/<TID_LOW>/<TID_HIGH>/data/**`00000001.sav`.
3. Compare the `agbsave.bin` counter against the SD slots' top and bottom counters.

Whichever slot is newer should match value with the `agbsave.bin`. 

(A) If the `agbsave.bin` and top slot counters are equal, increase the agbsave's counter by +1 and inject in the bottom slot.

(B) If the `agbsave.bin` and top slot are not equal, increase the agbsave's counter by +1 and inject in the top slot.

In the event this single slot injection somehow fails (maybe due to incompatible ARM7 registers??), that gba vc title should be able to load its last made savedata as a safety net.

PR #557, opened 2019-10-26
Change timeout on CardSPIWaitWriteEnd to improve responsiveness

dratini0:
Allows long waits for erase on things like Art Academy, but fails faster
elsewhere. In particular initialization fails almost instantly.


Wolfvak:
I'm trusting this has been well tested...

PR #569, opened 2019-11-27
fix #563

### LINKED ISSUE #563: [BUG] DISA/DIFF DataAbort Crash
when you mount arbitrary created DISA/DIFF and trying to write to it, GM9 crashes without writing anything.

(This was reported on GM9 Discord; escalating into issue)

aspargas2:
This was a very simple oversight that resulted in a stack-uninitialized pointer variable being used as if it was a pointer to a heap buffer.


aspargas2:
Apparently that pointer was being initialized to 0, upon further investigation. In any case, it was being left as something that wasn't the heap buffer it was supposed to be.

d0k3:
Merged, thankyou!

PR #570, opened 2019-11-28
add a9lh stage2 payload removal to the uninstall hax script

aspargas2:
This code will check if the hash of NAND sector 0x5BFFF matches the hash of NAND sector 0x5C000, and fill 0x89C00 bytes starting at sector 0x5BFFF with the byte at the beginning of sector 0x5BFFF if the hash does not match. Under unmodified circumstances, both of these sectors will be filled with 0xFF or 0x00. Sector 0x5C000 is where the stage2 payload gets written when installing a9lh, with a maximum size of 0x89A00 bytes. If this payload is not removed when uninstalling CFW, then attempting to reinstall it later using any method which involves b9stool will cause a false positive a9lh warning to be given.

I also changed a couple minor things to be consistent between the uninstall code for new models and old models.


d0k3:
Merged to staging, thank you!

PR #574, opened 2019-12-24
Disarm anti savegame restore (scripts).

TurdPooCharger:
I can also adjust the scripts where the user is explained what is the **00010011**/`00000000` and asked whether they would like to continue deleting it.

### DIFF
--- a/resources/gm9/scripts/GM9Megascript.gm9
+++ b/resources/gm9/scripts/GM9Megascript.gm9
@@ -167,7 +167,7 @@ end
 imgmount $[NANDBAK]
 inject -n I:/nand_minsize.bin E:/nand.bin@0
 imgumount
-rm -o -s 1:/data/$[EMUID0]/sysdata/00010011/00000000
+rm -o -s 4:/data/$[EMUID0]/sysdata/00010011/00000000
 echo "$[NANDBAK] successfully restored to EmuNAND."
 goto MainMenu_Restore_Options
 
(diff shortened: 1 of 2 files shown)


d0k3:
Okay, that's quite a few commits. Can I consider this finished and tested, @TurdPooCharger ?

TurdPooCharger:
I meant to push a single commit that edited both scripts at the same time. Edits were done directly on github to the `patch-1` forked repo and not on github desktop client. Online editing doesn't allow pushing a single commit with multiple files, and I couldn't be bother swapping back to my other SSD with the Linux setup that has the client as I felt the edits were simple enough. The last three commits were a mistake on my part due to forgetting emunand ctrnand uses `4:/` drive. 

Edits were done to these options. 

`GM9Megascript.gm9`
- **SysNAND Restore (full)**
- **SysNAND Restore (safe)**
- **EmuNAND Restore**
- **CTRNAND Transfer**

`NANDManager.gm9`
- **Y Restore EmuNAND**
- **L Restore SysNAND (Full)**
- **R Restore SysNAND (Safe)**
- **LEFT Transfer Emu to Sys**
- **RIGHT Transfer Sys to Emu**

The **Lazarus3DS** option found in `GM9Megascript.gm9` was left out because anyone resorting to that level of 3DS firmware repair is probably starting from scratch without preinstalled games to recover from.

Lastly, I did not personally test the individual options with the edits simply because I'm all too familiar with nand / firmware repair in GodMode9 by both manual and script methods. 

***

The commits as they stand will make the scripts function no differently than if they weren't included, except for the fact their respective **X:/data/XXXID0/sysdata/00010011/**`00000000` gets automatically deleted. 

If you rather allow the users using these scripts to decide for themselves whether or not they want to keep or delete that file, I can go back and readjust with `ask` command.

d0k3:
Okay, squashed and merged to staging. Thank you!

PR #576, opened 2019-12-31
minor changes to megascript luma3ds ctrnand setup

aspargas2:
add copying luma/payloads and remove the unneeded `-a` in the allow command because only lvl 1 perms are needed here


d0k3:
Merged to staging, thank you!

PR #577, opened 2020-01-06
fix config save dumping for citra in megascript

### LINKED ISSUE #566: [BUG] GM9Megascript/Dump_CITRA_Files dumps config save incorrectly
https://github.com/d0k3/GodMode9/blob/499e301b7b84c8e0b8b42a663bd2ed94d2ebcdec/resources/gm9/scripts/GM9Megascript.gm9#L933

Sorry if I read the script wrong, but it seems that dumping the config save is implemented as copying a region of data directly from the system save.

It is already known that the system file is in DISA format, while citra expects the format after unpacking the DISA file system. This is not as simple as copying one specific region of the file, and doing so can result in corrupted/inconsistent data. As DISA handling is already implemented, I request to correct this. If it is still too hard, this can simply be removed because citra nowadays mostly works without dumping this file (while having a higher chance to crash with an incorrectly dumped one).

aspargas2:
Fixes #566


aspargas2:
@wwylele Does this look good? I tested it on my system and the output is identical to extracting my config save with 3ds-save-tool.

wwylele:
Looks good to me.

PR #586, opened 2020-02-05
Support to variable sized tickets.

luigoalma:
Support to tickets of variable sizes added when processed individually or on ticket.db.
Lacking support on CIA building and mounting without the common ticket size as for currently, just added detection and error on building if found ticket is not of supported size.
Also a lot of indentation fixes. As well some memory related fix-ups.

A few notes:
- `Ticket` struct now has a content_index as a flexible array, should be used with malloc/calloc/realloc or a pointer cast to a predefined buffer length that can fit the needs of the ticket needed to load.
- `TicketMinimum` is available for the bare minimum ticket size, usable to check for it's real size, or content index is not required in full for a task.
- `TicketCommon` is currently for the CIA mount and build, with the common 0xAC context index size, while variable size isn't supported.
- Both a `TicketMinimum` pointer and a `TicketCommon` pointer can be safely casted to a `Ticket` pointer, as long for `TicketMinimum` the usage of this pointer doesn't exceed it's length, and `TicketCommon` has a properly set 0xAC content index size after load or fabrication.
- `TicketRightsCheck_InitContext` and `TicketRightsCheck_CheckIndex` expect ticket loaded completely in memory.
- `TicketRightsCheck_InitContext` can recycle the same context variable safely with a new ticket.
- `TicketRightsCheck_CheckIndex` is not to be called after used ticket on context as been unloaded from memory.


aspargas2:
Took a very brief look over the changes to my code and they look fine. I'll look at them in more detail whenever I get back to working on the virtual mounting thing; for right now I trust that they work fine. That code is unused at the moment, anyway.

d0k3:
Alright, I had a look - sorry this took so long, I've been very busy these days. Your changes look good to me. I'd rather avoid the pointers to pointers thing, cause it increses the risk of forgetting a `free` somewhere, but I understand this is required now with variable sized tickets.

Before I merge this, I'll wait until @aspargas2 had a closer look (no pressure, take your time). Maybe @Wolfvak can have a look, too, as this is a rather large one?

arm9/source/filesys/filetype.c
```diff
@@ -20,7 +20,7 @@ u64 IdentifyFileType(const char* path) {
     const u8 png_magic[] = { PNG_MAGIC };
 
     if (!path) return 0; // safety
-    u8 ALIGN(32) header[0x200]; // minimum required size
+    u8 ALIGN(32) header[0x2B8]; // minimum required size
```
Wolfvak: I'm not sure if having a buffer with a size that's not 32byte aligned is safe when using ALIGN(32) - if it's a minimum required, just align it to the next 32byte alignment.

luigoalma: Depends on the circumstances, if one's expecting the compiler to align up it's size too, leaving space in front, there's no guarantee here, and sizeof() still does give its normal size. If just expecting compiler to just align buffer position to expected alignment, then it should safe (assuming no issues on the compiler).
In this function, there should be no concern, but I can still change it if you prefer so.

Wolfvak: I've stopped trusting aligned buffers on GCC since the devs seem to change their mind on what exactly to assume every one or two major versions - a recent GCC upgrade broke aligned/packed buffers and debugging it took ages, and I'd rather not have something similar again.

I do expect it to pad the buffer size to the alignment, but it'd be better to explicitly do it in the code.

luigoalma: Yeah I get ya.
I'll change it then

d0k3:
@aspargas2 - I suggest we merge this and you give it another look when you add your future changes. Both, @Wolfvak and me already reviewed it, and if we don't merge it now we will be facing bad merge conflicts in the future.

aspargas2:
@d0k3 Yeah, definitely go ahead and merge if you've looked it over and it's fine.

PR #592, opened 2020-03-29
Always inject correct sighax signatures when installing b9s in the megascript

aspargas2:
Someone just bricked by accidentally using the dev b9s firm because there was no check for it and this was not being done. Please don't merge this yet as it has not been tested; I will undraft it once I get around to testing.


aspargas2:
Ok, tested now. Should be ready to merge, and hopefully save someone from bricking down the road.

PR #597, opened 2020-04-26
remove usm files in sd cleanup script

aspargas2:
The unSAFE_MODE exploit that will be soon added to the 3ds guide adds a few more files that can be removed.

### DIFF
--- a/resources/gm9/scripts/GM9Megascript.gm9
+++ b/resources/gm9/scripts/GM9Megascript.gm9
@@ -1175,6 +1175,9 @@ end
 if	not allow -a 0:/
  	goto MainMenu_Scripts_from_Plailect's_Guide
 end
+if      not allow -a A:/
+        goto MainMenu_Scripts_from_Plailect's_Guide
+end
 
 rm -o -s 0:/arm9.bin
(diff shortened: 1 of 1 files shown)


resources/gm9/scripts/GM9Megascript.gm9
```diff
@@ -1175,6 +1175,9 @@ end
 if	not allow -a 0:/
  	goto MainMenu_Scripts_from_Plailect's_Guide
 end
+if      not allow -a A:/
```
eku, line 1178: Indent?

aspargas2: Ugh, indent. That was a tab/space mixing oversight on my part. Thanks for pointing it out, but I don't think I'll bother with a whole new PR just to fix that because it's only a minor eye sore and the scripting language doesn't care about indents at all.

d0k3: @aspargas2 - maybe fix it locally and include the indent fix in the next script related pr you do?

aspargas2: Yeah, I'll do that.

PR #610, opened 2020-07-21
Allow installation of game files

### LINKED ISSUE #580: [FEATURE REQUEST] Add the ability to install carts and 3ds format game dumps to the SD card without converting to .cia first.
It would be nice to be able to install physical cart based games or 3ds format game dumps to the SD card without having to convert it to a .cia first.  Cartinstall can do this with physical carts but not 3ds files.

While it is possible to install cart based games or 3ds format dumps by converting them to a .cia file it is at least twice as slow and requires twice the free space on your SD card.  The cartinstall fork of gm9 can do this with physical carts but not .3ds files but presumably the same method could be made to work with them as well.

d0k3:
Should work for NCCH, NCSD, CIA, TMD from NUS/CDN and DSi eShop titles in NDS format.

This is not finished, and trying it may cause *anything* to happen. I warned you. Have fun.

When finished this will fix #580


d0k3:
Merged with master!

PR #617, opened 2020-08-04
add a needed delay for new 3ds consoles

Wolfvak:


arm11/source/system/sys.c
```diff
@@ -107,6 +107,10 @@ void SYS_CoreZeroInit(void)
 		mmuMapArea(0x20000000, 0x20000000, 128UL << 20, MMU_FLAGS(MMU_CACHE_WB, MMU_READ_WRITE, 1, 1));
 	}
 
+	if (SYS_IsNewConsole()) {
```
eku, line 110: Why is this needed?

PR #618, opened 2020-08-05
Finishing game installer functionality

d0k3:
This is a draft pull request, intended for testing and continued development (fixing bugs, adding missing feature) of the game image installing feature.

Here's a non-conclusive list of stuff still left to do:
* Identify and properly test all possible cases, fix any bug that comes up
* Implement cleanup on failed install (this needs uninstaller functionality first)
* Better handling for common issues (not enough space, title.db / import.db missing) with error messages
* Installer function for tickets(?)

Thanks go to @TurdPooCharger who has already started testing, looking forward to further bug reports.

PR #621, opened 2020-08-20
Finalize game installer functionality

d0k3:
This is a followup pull request to already merged #618. Based on the tests done previously by @TurdPooCharger (again, thanks a lot!), this is the current state:

**Presumably fixed**
Fixed, and tested at least once. Further testing makes sense here, cause we may still have missed something.
* bad CMD file without CMACs (see test 1): Should be fixed for good. Only system CMDs have no CMACs.
* `title.db` entry, bytes 0x0C...0x0 (see test 1): "title version" - takes over NCCH title version too for NCCH with DLP only (thanks @ihaveamac)
* cleanup of orphaned files from earlier installs (see test 2): Before an install, we wipe the `content/data` folder now. Existing saves with the correct size are left untouched.
* ticket device ID is always set to zero (see test 3): We keep the device ID if it matches the system now. It is zeroed only where required.
* savegames for TWL are not auto-created (see test 4): This is implemented now. The save files won't match system created ones exactly (in fact, only the first 0x200 byte should be identical for `public.sav` and `private.sav`), but it should still work. We are unsure if `banner.sav` also work. Further testing is required.
* `title.db` entry, bytes 0x20...0x23 (see test 5): "extdata id low" - this is now properly taken over

**Unfixed / Unfixable**
Stuff that we lack the knowledge for fixing. Some of these are "non-errors" and rather stuff that system / FBI does different from GM9. None of these has an effect on functionality.
* `title.db` entry, bytes 0x50...0x53 (see test 1): "mystery bytes" - we just write "GM9" there. Causes no problems at all. We'd very much like to do this proper, though.
* `title.db` entry, bytes 0x14...0x1D (see test 2): "TMD ID" and "CMD ID" - we always use 0 / 1 for this. This is not an error, but rather a slightly different approach. Won't be fixed.
* `title.db` entry, bytes 0x00...0x03 and 0x30...0x3F for `DS INTERNET` (see test 3) - we have no clue what happens here. Values seem random.
* Difference in ticket between CIA install and .3DS install (see test 5): No clue what happens here yet, but no effect on functionality. I think @aspargas2 may have a better understanding of how this happens than me.

**Other**
This category is for other stuff that was observed during the tests.
* Out of memory error when swapping out the `Nintendo 3DS` folder (see test 1): This can cause problems. We will research this, but it does not have high priority.
* Software keyboard does not work (see test 3): Fixed by @Wolfvak - thanks!

### DIFF
--- a/arm9/source/common/swkbd.c
+++ b/arm9/source/common/swkbd.c
@@ -136,7 +136,7 @@ static void DrawTextBox(const TouchBox* txtbox, const char* inputstr, const u32 
         (*scroll) ? '<' : '|',
         (inputstr_size > input_shown) ? input_shown : inputstr_size,
         (inputstr_size > input_shown) ? input_shown : inputstr_size,
-        inputstr + *scroll,
+        (*scroll > inputstr_size) ? "" : inputstr + *scroll,
         (inputstr_size > input_shown) ? 0 : input_shown - inputstr_size,
         (inputstr_size > input_shown) ? 0 : input_shown - inputstr_size,
         "",
--- a/arm9/source/game/ncch.c
+++ b/arm9/source/game/ncch.c
@@ -1,10 +1,7 @@
 #include "ncch.h"
-#include "support.h"
-#include "disadiff.h"
 #include "keydb.h"
 #include "aes.h"
 #include "sha.h"
-#include "ff.h"
 
 #define EXEFS_KEYID(name) (((strncmp(name, "banner", 8) == 0) || (strncmp(name, "icon", 8) == 0)) ? 0 : 1)
 
@@ -51,100 +48,6 @@ u32 GetNcchCtr(u8* ctr, NcchHeader* ncch, u8 section) {
     return 0;
 }
 
-u32 GetNcchSeed(u8* seed, NcchHeader* ncch) {
-    static u8 lseed[16+8] __attribute__((aligned(4))) = { 0 }; // seed plus title ID for easy validation
-    u64 titleId = ncch->programId;
-    u32 hash_seed = ncch->hash_seed;
-    u32 sha256sum[8];
-
-    memcpy(lseed+16, &(ncch->programId), 8);
-    sha_quick(sha256sum, lseed, 16 + 8, SHA256_MODE);
-    if (hash_seed == sha256sum[0]) {
-        memcpy(seed, lseed, 16);
-        return 0;
-    }
-
-    // setup a large enough buffer
-    u8* buffer = (u8*) malloc(max(STD_BUFFER_SIZE, SEEDSAVE_AREA_SIZE));
-    if (!buffer) return 1;
-
-    // try to grab the seed from NAND database
-    const char* nand_drv[] = {"1:", "4:"}; // SysNAND and EmuNAND
-    for (u32 i = 0; i < countof(nand_drv); i++) {
-        UINT btr = 0;
-        FIL file;
-        char path[128];
-
-        // grab the key Y from movable.sed
-        u8 movable_keyy[16];
-        snprintf(path, 128, "%s/private/movable.sed", nand_drv[i]);
-        if (f_open(&file, path, FA_READ | FA_OPEN_EXISTING) != FR_OK)
-            continue;
-        f_lseek(&file, 0x110);
-        f_read(&file, movable_keyy, 0x10, &btr);
-        f_close(&file);
-
-        // build the seed save path
-        sha_quick(sha256sum, movable_keyy, 0x10, SHA256_MODE);
-        snprintf(path, 128, "%s/data/%08lX%08lX%08lX%08lX/sysdata/0001000F/00000000",
-            nand_drv[i], sha256sum[0], sha256sum[1], sha256sum[2], sha256sum[3]);
-
-        // check seedsave for seed
-        u8* seeddb = buffer;
-        if (ReadDisaDiffIvfcLvl4(path, NULL, SEEDSAVE_AREA_OFFSET, SEEDSAVE_AREA_SIZE, seeddb) != SEEDSAVE_AREA_SIZE)
-            continue;
-
-        // search for the seed
-        for (u32 s = 0; s < SEEDSAVE_MAX_ENTRIES; s++) {
-            if (titleId != getle64(seeddb + (s*8))) continue;
-            memcpy(lseed, seeddb + (SEEDSAVE_MAX_ENTRIES*8) + (s*16), 16);
-            sha_quick(sha256sum, lseed, 16 + 8, SHA256_MODE);
-            if (hash_seed == sha256sum[0]) {
-                memcpy(seed, lseed, 16);
-                free(buffer);
-                return 0; // found!
-            }
-        }
-    }
-
-    // not found -> try seeddb.bin
-    SeedInfo* seeddb = (SeedInfo*) (void*) buffer;
-    size_t len = LoadSupportFile(SEEDDB_NAME, seeddb, STD_BUFFER_SIZE);
-    if (len && (seeddb->n_entries <= (len - 16) / 32)) { // check filesize / seeddb size
-        for (u32 s = 0; s < seeddb->n_entries; s++) {
-            if (titleId != seeddb->entries[s].titleId)
-                continue;
-            memcpy(lseed, seeddb->entries[s].seed, 16);
-            sha_quick(sha256sum, lseed, 16 + 8, SHA256_MODE);
-            if (hash_seed == sha256sum[0]) {
-                memcpy(seed, lseed, 16);
-                free(buffer);
-                return 0; // found!
-            }
-        }
-    }
-
-    // out of options -> failed!
-    free(buffer);
-    return 1;
-}
-
-u32 AddSeedToDb(SeedInfo* seed_info, SeedInfoEntry* seed_entry) {
-    if (!seed_entry) { // no seed entry -> reset database
-        memset(seed_info, 0, 16);
-        return 0;
-    }
-    // check if entry already in DB
-    u32 n_entries = seed_info->n_entries;
-    SeedInfoEntry* seed = seed_info->entries;
-    for (u32 i = 0; i < n_entries; i++, seed++)
-        if (seed->titleId == seed_entry->titleId) return 0;
-    // actually a new seed entry
-    memcpy(seed, seed_entry, sizeof(SeedInfoEntry));
-    seed_info->n_entries++;
-    return 0;
-}
-
 u32 SetNcchKey(NcchHeader* ncch, u16 crypto, u32 keyid) {
     u8 flags3 = (crypto >> 8) & 0xFF;
     u8 flags7 = crypto & 0xFF;
@@ -177,7 +80,7 @@ u32 SetNcchKey(NcchHeader* ncch, u16 crypto, u32 keyid) {
         if ((memcmp(lsignature, ncch->signature, 16) != 0) || (ltitleId != ncch->programId)) {
             u8 keydata[16+16] __attribute__((aligned(4)));
             memcpy(keydata, ncch->signature, 16);
-            if (GetNcchSeed(keydata + 16, ncch) != 0)
+            if (FindSeed(keydata + 16, ncch->programId, ncch->hash_seed) != 0)
                 return 1;
             sha_quick(seedkeyY, keydata, 32, SHA256_MODE);
             memcpy(lsignature, ncch->signature, 16);
@@ -355,5 +258,14 @@ u32 SetNcchSdFlag(void* data) { // data must be at least 0x600 byte and start wi
     exthdr->flag ^= (1<<1);
     sha_quick(ncch->hash_exthdr, &exthdr_dec, 0x400, SHA256_MODE);
 
+    return 0;
+}
+
+u32 SetupSystemForNcch(NcchHeader* ncch, bool to_emunand) {
+    u16 crypto = NCCH_GET_CRYPTO(ncch);
+    if ((crypto & 0x20) && // seed crypto
+        (SetupSeedSystemCrypto(ncch->programId, ncch->hash_seed, to_emunand) != 0) &&
+        (SetupSeedPrePurchase(ncch->programId, to_emunand) != 0))
+        return 1;
     return 0;
 }
--- a/arm9/source/game/ncch.h
+++ b/arm9/source/game/ncch.h
@@ -2,6 +2,7 @@
 
 #include "common.h"
 #include "exefs.h"
+#include "seedsave.h"
 
 #define NCCH_MEDIA_UNIT 0x200
 
@@ -16,13 +17,6 @@
 #define NCCH_NOCRYPTO  0x0004
 #define NCCH_STDCRYPTO 0x0000
 #define NCCH_GET_CRYPTO(ncch) (!NCCH_ENCRYPTED(ncch) ? NCCH_NOCRYPTO : (((ncch)->flags[3] << 8) | ((ncch)->flags[7]&(0x01|0x20))))
-
-#define SEEDDB_NAME         "seeddb.bin"
-#define SEEDDB_SIZE(sdb)    (16 + ((sdb)->n_entries * sizeof(SeedInfoEntry)))
-
-#define SEEDSAVE_MAX_ENTRIES  2000
-#define SEEDSAVE_AREA_OFFSET  0x4000
-#define SEEDSAVE_AREA_SIZE    (SEEDSAVE_MAX_ENTRIES * (8+16))
 
 // wrapper defines
 #define DecryptNcch(data, offset, size, ncch, exefs) CryptNcch(data, offset, size, ncch, exefs, NCCH_NOCRYPTO)
@@ -89,23 +83,10 @@ typedef struct {
     u8  hash_romfs[0x20];
 } __attribute__((packed, aligned(16))) NcchHeader;
 
-typedef struct {
-    u64 titleId;
-    u8 seed[16];
-    u8 reserved[8];
-} PACKED_STRUCT SeedInfoEntry;
-
-typedef struct {
-    u32 n_entries;
-    u8 padding[12];
-    SeedInfoEntry entries[256]; // this number is only a placeholder
-} PACKED_STRUCT SeedInfo;
-
 u32 ValidateNcchHeader(NcchHeader* header);
 u32 SetNcchKey(NcchHeader* ncch, u16 crypto, u32 keyid);
 u32 SetupNcchCrypto(NcchHeader* ncch, u16 crypt_to);
 u32 CryptNcch(void* data, u32 offset, u32 size, NcchHeader* ncch, ExeFsHeader* exefs, u16 crypto);
 u32 CryptNcchSequential(void* data, u32 offset, u32 size, u16 crypto);
 u32 SetNcchSdFlag(void* data);
-
-u32 AddSeedToDb(SeedInfo* seed_info, SeedInfoEntry* seed_entry);
+u32 SetupSystemForNcch(NcchHeader* ncch, bool to_emunand);
--- a/arm9/source/game/seedsave.c
+++ b/arm9/source/game/seedsave.c
@@ -1,0 +1,240 @@
+#include "seedsave.h"
+#include "support.h"
+#include "nandcmac.h"
+#include "sha.h"
+#include "ff.h"
+
+#define TITLETAG_MAX_ENTRIES  2000 // same as SEEDSAVE_MAX_ENTRIES
+#define TITLETAG_AREA_OFFSET  0x10000 // thanks @luigoalma
+
+// this structure is 0x80 bytes, thanks @luigoalma
+typedef struct {
+  char magic[4]; // "PREP" for prepurchase install. NIM excepts "PREP" to do seed downloads on the background.
+  // playable date parameters
+  // 2000-01-01 is a safe bet for a stub entry
+  s32 year;
+  u8 month;
+  u8 day;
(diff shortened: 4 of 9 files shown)


TurdPooCharger:
- `gameinstaller_finalize` compiled with last commit (f13b468).

## Test No. 9 - Install DLC CIAs.
I won't specify how many or which DLCs were involved in this test. Both the **FBI** versus **GM9** sets produced matching `*.app` contents with identical hashes. The only minor differences between the two were the ContentIDs of the `*.cmd` and their title entries. 
- **FBI** names a title's `*.cmd` by the total number of installed `*.app`. For example, if there are 100 (0x64) `*.app` files within the **A:/title/0004008c/<TID_LOW>/content/00000000**, the CMD will be named `00000064.cmd`.
- **GM9** always names the CMD as `00000001.cmd`.
- The [title entries](https://3dbrew.org/wiki/Title_Database#Title_Info_Entry) will likewise reflect the ContentID of their respect CMD at block 0x18–1C.
- Of the DLCs that were installed, one CIA had a TMD that specify a [Content Count](https://www.3dbrew.org/wiki/Title_metadata#Header) of 11 (**00 00 00 0B**). In actuality, there were only 10 `*.app` within the archive. After FBI installing this DLC, the output CMD was named `0000000a.cmd`. 
- The installed DLC with one less APP was checked with **TMD file options...** → **Verify file**. The result was **Verification failed** or a 'bad or incomplete dump'. I actually like that GodMode9 catches this outlier. It lets the user know whether his/her backups are intact. :+1: 

As was noted by @d0k3 in the [above reply](https://github.com/d0k3/GodMode9/pull/621#issue-471132810), the difference in CMD naming convention is not categorized as a bug. If there is a later change of mind or an exception made for DLCs to closer follow FBI/Dev Menu/custom-install, could the CMD's ContentID be copied from the CMD's final value for block 0x8–C?

## Test No. 10 - Can GameInstaller fix a fairly softbricked 3DS firmware? How does this compare to CTRTransfer?
This test is _brutal_, pitting GM9 against a noob's folly of messing with the NAND. Here's how it was conducted.

1. Download the `2.1.0-4U_ctrtransfer_o3ds.bin` image from [quantumcat1.github.io](https://quantumcat1.github.io/), and put this on the SD card. Insert the SD card to o3DSXL, and boot to GameInstaller.
2. Delete the **dbs** and **title** folders within **[1:] SYSNAND CTRNAND**.
3. Mount the 2.1.0 ctrtransfer image. Copy and paste its **dbs** and **title** folders to CTRNAND. Fix the six (6) `*.db` CMACs.
4. Backup the system titles into legit CIAs. There are 136 in total.
5. Safely eject the SD card. Press (START) to exit. Observe that the o3DSXL can boot to HOME Menu under firmware 2.1.0.
6. Return to GameInstaller. Except for `certs.db`, delete the other five (5) `*.db` files.
7. Press (START) to exit. Observe the o3DSXL now boots to `Black Screens of Death`.
8. Return to GameInstaller. Observe those previously deleted `*.db` files have been restored with dummy copies.
9. Delete the **title** folder. Press (START) to exit. Observe the luma error for missing NATIVE_FIRM,
> An error has occurred:
> Unable to mount CTRNAND or load the CTRNAND FIRM.
> Please use an external one.
> Press any button to shutdown

10. Return to GameInstaller. Install those CIAs created from step 4.
11. Press (START) to exit. Observe the o3DSXL boots to HOME Menu again.
12. (Attempt to) update to 11.13U: **System Settings** → **Other Settings** → **System Update**.

While my expectation of a successful outcome was one where the o3DSXL does update all the way, I wasn't going label this as a bug or hold this against GameInstaller if it didn't. Rest assure the o3DSXL updated to 11.13U. :+1: 

Some weird observations after the update.
- **Search for titles** found 160 titles/TMDs. A lot of the original versions for titles in **00040130** ([System Modules](https://www.3dbrew.org/wiki/Title_list#00040130_-_System_Modules)) were not deleted or replaced. System updating kept the old copies and installed the newest versions alongside those. These excess titles are effectively orphaned or leftovers. Why does this happens and how to fix this is not within the scoop of this PR.  
- Titles that were updated had their title entries adjusted with those mystery bytes (block 0x50–53) blanked out (**00 00 00 00**). This is strange because this doesn't happen when system updating from a higher (but not latest) version firmware.

I repeated the test again but this time with just the manual 2.1.0 ctrtransfer (skipping steps 1& 4–10). When I try to system update, this happens.
> An error has occurred.
> Unable to update.
> 
> The system will
> now be restarted.

The 3DS firmware semibricks/frankenfirm where **Search for titles** find 183 titles. The ones that fail in midupdate had lingering `*.ctx`.

The test is repeated once more where step 1 has substituted `11.5.0-38U_ctrtransfer_o3ds.bin` from [[3ds.hacks.guide] CTRTransfer](https://3ds.hacks.guide/ctrtransfer.html).
- The number of titles in 11.5U image also totaled 125.
- GameInstaller (all steps 1–12) was found to have 132 titles after the update.
- Manual transfer (only steps 2 & 11–12) managed to complete the system update. This too was found to have 132 titles.

Barring a messed up file system for the CTRNAND's FAT16 partition and noob stupidity limited to deleting/uninstalling the system titles (other nand files are not considered here), GameInstaller is a perfectly capable and acceptable tool/function when it comes to repairing the 3DS firmware matching and in rare instances, exceeding what standard CTRTransfer can do. :+1: 

## Test No. 11 - Homebrew system title(s)? SDK? CCI? CSU?
Yes.

**example homebrew system title**
- [FBI v2.6.0 (Full Dark Theme)](https://gbatemp.net/threads/release-fbi-dark-theme-and-theme-template-for-gimp.430810/page-2#post-8502738)

**dev titles**
- [[3dbrew.org] 3DS Development Unit Software](https://www.3dbrew.org/wiki/3DS_Development_Unit_Software)

GameInstaller can install other, less common NCSD and NCCH containers. Whether a specific title launches and runs is another story.

***

The write-up for Test No. 12 will be present tomorrow. After starting, stopping, and deleting the first sentence several times, this one is proving rather difficult in how to best present the findings.

TurdPooCharger:
## Test No. 12 - Are bad dumps/corrupt CIAs rejected?
Using the CIA version of FBI v2.6.0, I hex edit an arbitrary location at block 0x3C76–3C78 from (**00 00 00**) → (**42 41 43**) or 'BAD'. This region is found within the `extheader.bin` for the title's `EE360966.app`.
- The SHA-256 for the original `*.app`: 2603335FD7BFDD13C92C17A9653A597651DEDD55A7CC3355281C66D24BEA5976
- The hash for the edited `*.app`: 937545E80AE0BEEF3DDF3DA7E89A5E48F912BDFF2E9042492925F80CD3822CCA

I rename this corrupted CIA to `FBI (bad).cia`. As can be expected when checking this file,
- GodMode9 → `FBI (bad).cia` → **CIA image options...** → **CIA checker tool** → **<A> yes** → Result: **Verification failed**
- ... → **Verify file** → Result: **Verification failed**

, this is bad dump.

I attempt installing `FBI (bad).cia` with the 3DSX version of FBI and DevMenu. The results are as follows.

FBI_DM_#12

Next, I try installing `FBI (bad).cia` with custom-install... The PC program allows the installation to go through. I eject the SD card, reinsert it to the o3DSXL, and check if the installed title is still considered a bad dump.
- GodMode9 → **[A:] SYSNAND SD** → hold (R)-shoulder and press (A) → **Search for titles** → `... FBI (CTR-P-CFBI) (W).tmd` → **TMD file options...** → **Verify file** → Result: **Verification failed**

The **Nintendo 3DS** folder is set aside and renamed as **Nintendo 3DS (custom-install)**. A fresh copy of the **Nintendo 3DS (reference)** is pulled and renamed as the replacement **Nintendo 3DS**. The folder will later be referred to as **Nintendo 3DS (GameInstaller)**.

***

**BUG**
Just like @ihaveamac's custom-install, I was able to install `FBI (bad).cia` with GameInstaller. Some people may have a different opinion than I do on this and in certain exceptions (like the above Test No. 9 for the DLC with the missing `*.app`), I find it rather more important that the CIAs are hashed and verified against the TMD's [content chunk records](https://www.3dbrew.org/wiki/Title_metadata#Content_chunk_records) than that they can be installed at all. If a CIA in question is found to be a bad dump, rejected installing it. Save everyone the headache in the long run trying to troubleshoot why their games crash or aren't stable. 

### SERIOUS BUG
After installing the CIA to **Nintendo 3DS (GameInstaller)**, I checked the title. What I saw was very troubling.
- Result: **Verification success**

The TMD's chunk record had the hash data field adjusted to the corrupt APP's SHA-256: 937545E80AE0BEEF3DDF3DA7E89A5E48F912BDFF2E9042492925F80CD3822CCA

This bad installation of FBI could be dumped back into a CIA again where **CIA checker tool** and **Verify file** stated it to be **Verification: passed** and **Verification success**. 

In certain instances like encrypting or decrypting CIAs or romhacking, I can see where it would be necessary to edit those hash records to reflect the (proper) changes that have occurred to the APPs. 

No exceptions should be made for corrupt titles! All this leads to is propagating bad dumps disguised as good ones.

### (Another) SERIOUS BUG
This bug is not directly related to GameInstaller, but finding this broke the preservationist's heart within me. :-(

I image mounted an original intact copy of `FBI.cia`, **[CLIPBOARD]** selected the `tmd.bin`, navigated to the **A:/title/0004000/0f800100/content** directory where bad FBI is installed to, pressed (A) on the `00000000.tmd`, and **Inject data @offset** → **|00000000|**. 

After overwriting the `*.tmd`, I checked if the installation was now seen as bad.
- **TMD file options...*** → **Verify file** → Result: **Verification failed**

This is a very handy feature to have when one has their library corrupted by a fake SD card and would like to salvage any titles that are still intact after migrating to a replacement card.

Despite the failed verification, I am able to build a CIA from this corrupted title. That CIA had its TMD chunk record's hash adjusted to the bad APP's hash, giving a misleading impression that the dump is intact. The ability to back up bad titles into CIAs has been present in older versions of GodMode9.

TurdPooCharger:
Except for the possibility of (Test No. 13) Citra decrypted `*.3ds` rom set to the wrong crypto flag and (Test No. 14) SD card that doesn't have enough free space, I have no other scenarios in mind to try out.

If anyone has any suggestions or believe something is missed, any further testing (other than maybe 13&14) will focus on revisiting bugs from earlier.

d0k3:
Sorry about the radio silence! I'm on vacation right now, I don't even have any dev hardware with me (will be back in 2.5 weeks). Thanks a lot for all the testing! I'll get to all of it eventually.

For now, I wonder about the seeds. Isn't that something the 3DS OS should take care of? Ie, aren't seeds autodownloaded when the system is connected to the internet?

I'm also full aware we can't rely on Nintendo keeping their servers running for eternity (see: Wii servers and Miiverse), so we may need to find a solution one way or another.

luigoalma:
> For now, I wonder about the seeds. Isn't that something the 3DS OS should take care of? Ie, aren't seeds autodownloaded when the system is connected to the internet?

I can something about that.
NIM handles downloading seeds outside the eshop (can't confirm if eshop itself does that too when opened or if or when it prompts NIM).

It downloads for a selected title id and country id at nim:u IPC 0x002D by registering for download on the background daemon thread or nim:s IPC 0x005F the same way on the spot by demand.

Also nim:u IPC 0x002E by registering for download on the background daemon, downloading all seeds **only** registered on FS's TITLETAG file (stored in sysdata 0x0001000F) as a prepurchase install (NIM handles, registers and updates these also). Or while the console is sleep mode (still for prepurchase registered titles only) after being awake by SRV notification 0x107 and PTM, if there are no background downloads left waiting to be performed.

I've not seen seed autodownloading asides what is registered as a prepurchase just from the home menu alone (if someone wants to test that further though, feel free).

If you don't install in the seed itself to SEEDDB, you may install the TITLETAG entry, it will make it look like a prepurchase install too on the home menu, so it will download when user tries to open it, or user leave console in sleep mode long enough.

d0k3:
Thanks a lot, @luigoalma, I wasn't aware of this. I guess we need to provide some way of injecting seeds in the long run. Servers won't run forever.

Is there a complete listing of which games use seed crypto? We're mainly interested in the numbers right now.

luigoalma:
It was mentioned here before, ihaveamac's [SeedDB List](https://github.com/ihaveamac/3DS-rom-tools/wiki/SeedDB-list).

TurdPooCharger:
I can comparison files of the **0001000f/**`00000000` of before and after importing the seeds for example games.

TurdPooCharger:
Sorry for ghosting on you guys this past month.
<details>
<summary>[Spoiler] not relevant to this commit. What I've been up to.</summary>
A bunch of random stuff came up that dampened my time, energy, and mood spent online. 

- Despite wearing gloves, KN95 + surgical masks, chemical splash goggles, separate set of PPE style clothing [hoodie, khaki pants, shoes, socks, shirt, etc.] and having a +70% alcohol sanitizer spray bottle carried at all times whenever going out to public for groceries and food pick-up, I managed to get the _tiniest_ of sickness and went into self-quarantine.

- Jerry rigged my 12 years old Samsung laundry machine. The washer tub had a terrible balance where it would violently shake the house during a fast spin cycle. Not even a new set of shock absorbers fixed the problem. Because my family's broke as shit and we didn't feel like replacing an otherwise perfectly fine washer, I decided to put my engineering degree to good use by retrofitting the washer with moar dampers! Instead of 4 shocks found on a typical washer, mine is rocking 8. :-) Take that planned obsolete!

moar_dampers

- Sold off two of our clunker cars. Fixed a third car with a snapped serpentine belt that was way past overdue on its 100K miles change interval. Opps on keeping up with basic maintenance. Looking to trade in our 4th car as down payment for a Japan-built 2021 Toyota Corolla SE. 

- This Atlantic hurricane season has been waaay more active than other years and has made us reconsider getting home insurance. My Floridian house, made of concrete walls and concrete tiled roof, has survived so far. Luck will run out sooner or later... :-/ Blah, the joy of wading through scheduling appointments for home inspections and paperwork. Fun stuff.
</details>

***

Attached in this ZIP are before and after changes for the **0001000f/**`00000000`, their `partitionA.bin`, and the individual seed DAT files for the below example games.
- [import seed stuff.zip](https://github.com/d0k3/GodMode9/files/5321000/import.seed.stuff.zip)

Order of seeds imported.
**0 - No imported seeds (empty).**
**1 - 1st game:** _Pokémon - Gold Version VC (USA)_ `(0004000000172600)`
**2 - 2nd game:** _Bloodstained: Curse of the Bloodmoon (JPN)_ `(00040000001D3A00)`
**3 - 3rd game:** _Pokémon - Silver Version VC (USA)_ `(0004000000172700)`

***

Here are the data fields I could gather for the `partitionA.bin`.

- 0x3004–???? = The counter value for the number of seeds contained within. I don't know which direction it rolls over (towards 0x3003 or 0x3005?) when you have more than 255 (FF) seeds. I'll see if I can mass import 255 seeds through a custom gm9 script, import one more seed in FBI, and then report back which way it goes.

TitleID of the games (endian) in the order they were imported.
- 0x4000–4007 = 1st game.
- 0x4008–400F = 2nd game.
- 0x3010–4017 = 3rd game.
.
.
etc.

Seeds of the games in the order they were imported.
- 0x7E80-7E8F = 1st game.
- 0x7E90-7E9F = 2nd game.
- 0x7EA0-7EAF = 3rd game.
.
.
etc.

***

For my 4th attempt, I tried to import the seed for _Pokémon - Gold Version_ again. No changes observed for the **0001000f/**`00000000`'s `partitionA.bin`; FBI does not allow you import redundant copies of the same seed.

d0k3:
Alright, sorry bout the radio silence, too. I've been pretty busy these last few weeks, slowly finding my way back into GM9 development. Thanks a lot for all of your help! I'll try to review everything that was tested so far and will make a new list during the next few days.

As for unrelated stuff, I guess there is no way to escape the common cold, no amount of desinfectant, gloves, masks will achieve that for us. Hope all is well! I like what you did to that washing machine, too.

d0k3:
Alright, after going through each test, here's my summary.

**1. Is that stuff fixed for good?**
I'm just taking over the list from the op here. I *think* everything here is fixed, but maybe @TurdPooCharger can have a second look.

* bad CMD file without CMACs (see test 1): Should be fixed for good. Only system CMDs have no CMACs.
* `title.db` entry, bytes 0x0C...0x0 (see test 1): "title version" - takes over NCCH title version too for NCCH with DLP only (thanks @ihaveamac)
* cleanup of orphaned files from earlier installs (see test 2): Before an install, we wipe the `content/data` folder now. Existing saves with the correct size are left untouched.
* ticket device ID is always set to zero (see test 3): We keep the device ID if it matches the system now. It is zeroed only where required.
* savegames for TWL are not auto-created (see test 4): This is implemented now. The save files won't match system created ones exactly (in fact, only the first 0x200 byte should be identical for `public.sav` and `private.sav`), but it should still work. We are unsure if `banner.sav` also work. Further testing is required here. Can you help @TurdPooCharger ?
* `title.db` entry, bytes 0x20...0x23 (see test 5): "extdata id low" - this is now properly taken over

**2. Stuff that may need further looking into (not high priority)**
Also a copy from the op. Nothing here has high priority. Just a quick reminder.

* `title.db` entry, bytes 0x00...0x03 and 0x30...0x3F for `DS INTERNET` (see test 3) - we have no clue what happens here. Values seem random.
* Difference in ticket between CIA install and .3DS install (see test 5): No clue what happens here yet, but no effect on functionality. Maybe @aspargas2 has a better understanding of how this happens than me.

**3. Stuff that must be looked into**
That's the stuff from tests #6 and upwards (I'm glad GM9 passed at least a few of these tests). I need to have a closer look here, and there's also some open questions (which I'll add to the list).

* installing games and handling seeds (see test 7): I think GM9 should at least import seeds from a user provided seeddb.bin. That should solve this issue for good. When the required seed is not available, we can at least try to install the TITLETAG entry (thanks @luigoalma), so the 3DS OS can look for the seed itself. @TurdPooCharger's pull request is also a great way of approaching this issue.
* update CIAs creating save files (see test 8): Some update CIAs have the save flag set, thus GM9 creates a save file for them (which it should not). I need ideas on how to solve to solve this. Should we disable generating save games for update CIAs in general? And can we properly recognise update CIAs?
* installing N3DS exclusives on O3DS (see test 8): I personally don't think that's an issue. GM9 does what the user wants after all. I'd be okay with putting out a warning (if we're able to recognise N3DS exclusives properly), but I don't think it's required.
* leftovers when installing system titles (see test 10): Oooof. Tbh I'm not really sure I understand what happens there. You were able to restore the system titles from GM9, then, after the updated, you noticed there were some leftover titles (older versions), right? There may actually be no solution to this, as the problem seems to be in the 3DS OS update and we do not completely know how this works. May be related to TMD numbering (which defaults to 1 in GM9). Can you give some more info, @TurdPooCharger ? Do you think this needs work?
* handling corrupt game images (see test 12): I agree, we at least need to make the user aware of corrupt game data. I'll look into it.

luigoalma:
Also, I should note as well. `SEEDDB` cap is 2000 seeds, in system modules (like NIM processing lists) and physically in space too. 0x3000 is where file *usually* starts on the `partitionA.bin` Inner FAT, although it can be elsewhere, its a filesystem.
`SEEDDB` is exactly 0xCB80 (52096) bytes in length.

Using offsets relative to `SEEDDB`, not `partitionA.bin`, based on information given by @TurdPooCharger and double checked by making a few tests and copying out `SEEDDB` with FBI:
- The first 0x1000 bytes are unknown (usually 0?) except at offset 0x4, which has a little endian u16 or u32 (unknown which, could be u16 since its big enough for 2000, that requires checking how FS reads it to know which data type).
- At 0x1000, title id list starts, fits 2000 little endian `u64` entries. This is 0x3E80 (16000) bytes.
- At 0x4E80, seed list starts, fits 2000 `u8[16]` entries. This is 0x7D00 (32000) bytes.

Might as well also tell the structure of `TITLETAG` file on the same Inner FAT too.
It is exactly 0x43680 (276096) bytes in size.
Follows the same format as `SEEDDB`, 2000 entries as well, except:
- At 0x4E80, title tag list start, fits 2000 titletag structure entries (0x80 bytes each). This is 0x3E800 (256000) bytes.

The structure of a title tag entry goes as it follows:
```c
// This structure is 0x80 bytes
typedef struct {
  char magic[4]; // "PREP" for prepurchase install. NIM excepts "PREP" to do seed downloads on the background.
  // playable date parameters
  // 2000-01-01 is a safe bet for a stub entry
  s32 year;
  u8 month;
  u8 day;
  u16 country_code; // enum list of values, this will affect seed downloading, just requires at least one valid enum value. 1 == Japan, it's enough.
  // everything after this point can be 0 padded
  u32 seed_status; // 0 == not tried, 1 == last attempt failed, 2 == seed downloaded successfully
  Result seed_result; // result related to last download attempt
  s32 seed_support_error_code; // support code derived from the result code 
  // after this point, all is unused or padding. NIM wont use or access this at all.
  // It's memset to 0 by NIM
  u8 unknown_filling[0x68];
} FS_TitleTagEntry;
```

d0k3:
Thanks @luigoalma - just the info I need when I need it!

d0k3:
@luigoalma - can you also tell me to usual start of the TITLETAG file inside partitionA?

luigoalma:
TITLETAG I believe, usually, 0x10000.
These offsets are if given the fact the user or homebrew did not try to delete and restore SEEDDB and TITLETAG files, or tried to do some other misc file adding or removable into syssave 0x1000f.

d0k3:
The last few commits should handle all the issues found in test 7 and test 12.

d0k3:
Any ideas about update CIAs creating save files? (see test 8)

d0k3:
Alright, as @TurdPooCharger will be unavailable for a while, I merged this to master. The remaining issues and tests will be handled in an issue.

PR #623, opened 2020-08-25
Fix GCC warning, snprintf limit too long

gemarcano:
- One snprintf in arm9/source/godmode.c call had a limit that was past
   the size of the datestr variable it was writing into. Fixed to match
   the size of the variable.


Wolfvak:
thanks!

PR #625, opened 2020-08-25
MCU Event implementation

Wolfvak:
add an atomic MCU event system to be used by the GFX code (and maybe anything else that could need it)

also refactors a bunch of other code that should've been cleaned up and simplified long ago


Wolfvak:
Alright, this PR adds an "event interface" for the ARM11 and changes most of how it works - instead of having an interrupt do its work (in a non-interruptible context) it now simply marks itself as "pending" in software.

The main PXI loop (and other pieces of code) can then check whether there are pending interrupts and act upon them. Although we lose in timing predictability, we gain in stability because all hardware management is done in an interruptible system context: for example, now we don't have to worry about a PXI call to read/write I2C interfering with the same call happening on a VBlank handler.

Although there's still a few more features left to implement/things to fix, this PR is now ready for testing and review.

d0k3:
Merged to current master, merge conflicts fixed

PR #634, opened 2020-10-05
Add new function: import all seeds.

TurdPooCharger:
Set it and forget it.


luigoalma:
I feel like this could be done a bit more delicately. `0001000f/00000000` is a DISA wrapper with a savegame file in it, given the SAVE header. GM9 can mount the DISA wrapper but not the Inner FAT savegame yet. Although most systems will have `SEEDDB` and `TITLETAG` positioned about equally the same place, its not strictly set to be always the same place. With the right type of tampering and `SEEDDB` file from the save is elsewhere.
Although be it most users likely wouldn't tamper with the save file.
But just putting it out there

TurdPooCharger:
@luigoalma, I see the `SEEDDB` at 0x2034–203A and the `TITLETAG` at 0x2064–206C in the `partitionA.bin` on my n3DSXL and o3DSXL,. Are you saying that the data fields around this area can shift from those locations? I've seen this occurrence before in the **00010017**/`00000000` (`configsave.bin`) where the offsets of the user info will depend on which 3DS firmware version the user created his/her profile.

Or are you referring to the effect where the 3DS falls back on the second partition `partitionB.bin`(?) when the active partition (`partitionA.bin`) is no longer valid from due to broken [chain of trust](https://www.3dbrew.org/wiki/DISA_and_DIFF#Chain_of_trust)?

As for the PR itself, the proposed changes does not edit anything before 0x3000 in the `seedsave.bin`'s `partitionA.bin` so the `SEEDDB` and `TITLETAG` stuff around 0x2000-208F are left alone.

Do you if the initial entries for the first TitleID and its seed is written to elsewhere than 0x4000 (TitleID) and 0x&e80 (seed0?

luigoalma:
It's essentially a filesystem. Fuzz around the file tree with just FS on userland and things can out of place than expected.
Not talking about partitionB.bin, this one happens to not have it.
Most users will have it the same place, but it's doing a guess.

d0k3:
I guess I'll wait a bit before I merge this. Let me know when it's ready, @TurdPooCharger !

d0k3:
Okay, a new commit allows handling seeds from `seeddb.bin` now when installing titles: https://github.com/d0k3/GodMode9/pull/621/commits/8e3a5b496af7e1d91c4ca7e11192800bf3658416

So, do we still need seed importer functionality in the megascript? If we do I at least suggest we get rid of the hardcoding. I can provide a scripting function that would import the seeds from `seeddb.bin` in that case.

TurdPooCharger:
Back from a long break.

I no longer believe this mass importing seeds method is necessary.

I opened this PR as a backup plan in case adding a built-in import seed functionality in GodMode9 never materialized.

After @luigoalma shared that key tidbit but before the break, I was in the beginning of research studying differences in the `seedsave.bin` between initial 9.6 and latest firmware versions. System title CIAs were being collected for the eventual downgrade and incremental updates. This was in hopes of finding all the possibility offsets of the `SEEDDB` and `TITLETAG`.

If there wasn't an exact pattern locating those stuff around the 0x2000–2100 area, one approach I had in mind to remedy this would have been to blank them out and hard code inject them back at the previous mentioned typical 0x2034–203A and 0x2064–206C.

There are two potential issues to this mass import idea.
- (1) Any leftover or new seeds added to ihaveamac's updated `seeddb.bin` would mean having to go back and manually edit the script to include those stragglers. This may or may not be a problem depending if Nintendo eShop(s) of JPN, EUR, USA, etc. are still releasing new digital 3DS titles.

- (2) After importing all the seeds, I noticed that my/cousin's o3DSXL when booting to HOME Menu that the screens tend to stay black with background music longer than usual, like up to 3-5 secs before the screens load up. This doesn't always happen, but this may lead users with a good eye to worry and think the slower loading than usual could be something wrong with their 2DS/3DS.

I'll close/rescind this PR once I get around to testing commit `8e3a5b4`.

d0k3:
Don't want to nag, @TurdPooCharger. As I'm going through the repo right now, I just wanted to give you a quick reminder that you wanted to close this one once `8e3a5b4` is tested.

TurdPooCharger:
@d0k3, no worries. I needed the reminder. Sorry about completely dropping the ball bug testing GM9/GameInstaller. I'll close this PR and report findings for seed stuff at the other issue.

Issue #457, opened 2019-02-10
Cannot dump Dsi dev carts


mariomadproductions:
Someone I know is trying to dump a DSi dev flash cart. But there is a "Cart Init Failed!" error message when he tries to dump it using GodMode9 1.7.1. Is there any more detailed debug info I can provide?

legoj15:
I'm going to go out on a limb and say if it is true that dev DSi carts are encrypted for dev DSi hardware, then they need dev DSi keys, something that a retail 3DS (or even dev 3DS) probably doesn't have. I could easily be 100% wrong, but I'd imagine that there'd be no reason to have DSi dev keys on a 3DS.

mariomadproductions:
Possibly. This could be something that could be put in the support files folder.
But I wonder why this would cause a cart init failure. Maybe failure to decrypt the ROM causes this error, even if technically the cart init worked?

legoj15:
Oh you may be right about the support folder thing. I think the explanation that the cart did init but failed to decrypt seems logical, but only @d0k3 would really know.

GerbilSoft:
https://problemkaputt.de/gbatek.htm#dsencryptionbygamecodeidcodekey1
https://problemkaputt.de/gbatek.htm#dsicartridgeheader

DS and DSi cartridges are initialized using commands 0x3C (NTR) and 0x3D (TWL), respectively. GBATEK lists three sets of Blowfish keys for KEY1 initialization:
* NDS (ARM7 ROM): 99 D5 20 5F...
* DSi (ARM7 ROM): 59 AA 56 8E...
* DSi Debug (stored in launcher): 69 63 52 05...

The DSi Debug key is used when SCFG_OP is non-zero, i.e. running on debug hardware.

There's also a debug-specific modcrypt key for the DSi secure area. This is indicated in the ROM header by setting bit 3 of 0x1C. When set, the modcrypt key is set to the first 16 bytes of the cartridge header. I'm not sure if GM9's DSi dumping functionality decrypts the modcrypt area or leaves it intact. Note that decrypting the DSi secure area is not required for cartridge initialization, just the Blowfish key.

d0k3:
[...]

PR #645, opened 2020-11-13
Fix dumping of TWL ROM images flashed to NTR dev carts

GerbilSoft:
We've encountered a few NTR dev cartridges that have TWL-enhanced ROM images flashed to them. These cartridges work fine on a Nintendo DS, and on some 3DS units. Attempting to dump them with GodMode9 (and wooddumper) fails because the ROM header indicates that the cartridge is TWL-enhanced, but the NTR dev cart doesn't support the TWL secure area init command (0x3D), so TWL key exchange fails.

This PR checks the TWL flag in the chip ID. If it's not present, then it won't attempt to use TWL secure area init, even if the TWL flag is set in the ROM header.

We're also using the ROM size value in the chip ID now for the full cartridge size if it's present. If not, it falls back to using the ROM header value.

### DIFF
--- a/arm9/source/gamecart/gamecart.c
+++ b/arm9/source/gamecart/gamecart.c
@@ -240,6 +240,20 @@ u32 ReadCartPrivateHeader(void* buffer, u64 offset, u64 count, CartData* cdata) 
     return 0;
 }
 
+u32 ReadCartId(u8* buffer, u64 offset, u64 count, CartData* cdata) {
+    u8 ownBuf[GAMECART_ID_SIZE] = { 0 };
+    u32 id;
+    u8 sReg;
+    if (offset >= GAMECART_ID_SIZE) return 1;
+    if (offset + count > GAMECART_ID_SIZE) count = GAMECART_ID_SIZE - offset;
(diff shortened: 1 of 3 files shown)


d0k3:
Okay, this already looks pretty good! It may also fix #457. Will this have any adverse effects on carts that previously worked?

d0k3:
Test build here:
https://f.secretalgorithm.com/SPO0o/godmode9-v1.9.2pre1-21-titleman-devcartdump.zip

GerbilSoft:
This shouldn't affect other carts that worked before:
* NTR carts usually don't have the TWL flag set in the chip ID. Kirby Mass Attack does, but it doesn't have the TWL flag set in the header, so it'll still be handled as NTR.
* TWL carts *do* have the TWL flag set in the chip ID.

I don't think it'll fix dumping TWL dev carts. That requires using the TWL secure init command, but with a different Blowfish key. I did extract the Blowfish key, but I'm not sure if it can be included in GM9 directly, or if there's any way to properly distinguish between retail and dev carts.

d0k3:
Partially merged. The gamecart ID will be added to the filename, thus the additional file is not required. Everything else is included.

PR #662, opened 2020-12-17
Fix uninitialized variable on unaligned SPI Xfer

Wolfvak:
We use a stack-allocated u32 to store a temporary word that gets memcpy'd from a potentially unaligned buffer, but the size of the copy could be less than 4 bytes, therefore leaving garbage in the upper bits of said word. This fixes CODEC in Corgi3DS.


d0k3:
Merged

Issue #553, opened 2019-10-20
[BUG] Non-working carts with the new save management


d0k3:
Quoting @wwylele here:
> Hi, I saw that godmode9 added cart save dumping. I tried it, but only 1 of 7 of my carts got the save recognized and dumped. For others the save file just doesn't display in the cart drive. Do you know what could be happening here?
> These are the games I tried (all Japanese version, on both Japanese o3ds and n3ds):
> mario 3d land
> new super mario bros 2
> pokemon y
> pokemon alpha sapphire 
> pokemon mystery dungeon: gates to infinity
> Kirby triple deluxe
> pokemon super mystery dungeon (This one succeeded)

Can't do much myself, we'll need your help @dratini0 - got any clue? And, @wwylele, can you add some info? Did these carts simply not display their saves in the gamecart drive?

wwylele:
Right, they simply didn't display their saves in the gamecart drive.

Wolfvak:
card2 games aren't supported at all, dunno if those are.

wwylele:
Among my list only the two big pokemon games (Y & AS) are card2. Others are card1

dratini0:
Hmm, memory serves me right ORAS and gates to infinity are card2 as well, but I might be wrong.  In any case, I will have code that can ascertain this ASAP.

In the meantime you can tell for sure by looking into the header for the 3ds file in the hex editor, finding the line that says 188 and checking the sixth byte (offset 0x18d). If it is 01, it is a card1, if it is 02, it is a card2. I am not sure what 00 or 03 means.

wwylele:
> Among my list only the two big pokemon games (Y & AS) are card2. Others are card1

I posted this right after I checked my games using ctrtool. I assume the card type reported there is correct?

dratini0:
Oh, sorry. Probably it did the same thing as I was suggesting, making me a smartass. In this case, I have no clue what is going on.

dratini0:
My best guess at this point is buying samples and opening them up and/or looking at the transactions with a logic analyzer. In any case, thank you for pointing this out - this will take me some time.

Ammako:
> In the meantime you can tell for sure by looking into the header for the 3ds file in the hex editor, finding the line that says 188 and checking the sixth byte (offset 0x18d). If it is 01, it is a card1, if it is 02, it is a card2. I am not sure what 00 or 03 means.

Don't need to, you can just look at the serial on the label. EXXX are card2.

[...]

PR #670, opened 2021-01-31
Save management: Detect save size by the last byte of JEDEC id, if possible

dratini0:
This makes progress on #553, but not complete closure. Not by a long way...

The long and short of this is that this commit now only compares the first two bytes of the JEDEC ID of the flash chip when looking it up. When looking up the size, if the lookup table has a size, it uses that, otherwise it checks the last byte of the JEDEC ID. Additionally, the auto-detection for type is now skipped for card1 3DS cards, and all the flash chips are handled the same way.

I have also tried another approach on the branch ctr-generic-flash, in which I read from the ExHeader instead. I think I prefer this approach, as this doesn't involve going deep into the cartridge header. Then again, that might be needed for card2 games. (Unless I can just assume the save area just stretches to the end of the cartridge?...)

@wwylele , Thank you again for the pointer! Would you mind trying this one out, especially the writing, if you feel brave enough? Reading is practically guaranteed to work, and if the datasheets are to be believed, the writing will be just fine, too. (It now uses the same parameters it uses for the Macronix chips)


d0k3:
Glad to see progress on this one! I had a look through the code, looks good to me. I guess I'll still wait with merging until @wwylele comments, as there may be more changes needed.

About the card2 save size - my cart reading code assumes the card2 save actually stretches until the end of the cart storage. For cart reading, that's a save assumption, as there can't be anything else but save data or padding there, and we don't want save data in cart dumps. That may not be the case for save writing, as I'm not sure about the size of the writable area. I don't know about any info inside the cart header that would specify the actual size of the header, but I may also have missed it.

dratini0:
> I don't know about any info inside the cart header that would specify the actual size of the header, but I may also have missed it.

I will assume you mean "actual size of the save" here.

I have some code that parses the header for the save size on my other branch here: https://github.com/dratini0/GodMode9/blob/ctr-generic-flash/arm9/source/gamecart/gamecart.c#L96

It boils down to the following:
1. Read the first two sectors of the first partition on the cartridge (NCCH and first sector of ExHeader)
2. Decrypt the ExHeader using the NCCH
3. Look up offset 0x1C0 in the ExHeader

It does seem to arrive at the right result, but it just seems like a very roundabout way to do it for SPI flash based saves.

profi200:
The only way to get the CARD2 save size (for 3DS cards) is via the exheader. The offset where the save is stored is in this header: https://www.3dbrew.org/wiki/NCSD#Card_Info_Header

d0k3:
@dratini0 - I assume this is ready to be merged (also in regards to the coming release, as mentioned in that other pr discussion)?

dratini0:
I am quite happy with it. That said I haven't tested it with any of the `0x6226XX` games, because I don't have any and I am quite concerned that they don't keep the save chip consistent between production runs of the same cartridge. The best I can say that it still works with the `0xC222XX` cartridges I have.

dratini0:
I got my hands on a bootleg copy of Pokemon Diamond, and it turns out that that has a previously unsupported type of flash chip. (ST M25PE40VP) So I added support for that too (not that bootlegs working/not working for any particular purpose isn't a bit of a crapshoot anyway). It does look pretty fast and versatile to me.

dratini0:
I have rebased this as per the discussion on #672

dratini0:
I pushed my changes to fix #690 here, as I don't feel like starting a zillion merge requests and rebasing them again.

d0k3:
Alright, that means it's good to merge now?

dratini0:
Thank you. Yes, that should be fine

d0k3:
Merged with current master. Thanks a lot, @dratini0 !

PR #672, opened 2021-02-04
Save management: Card2 reading

dratini0:
This is some initial support for reading card2 saves. At the moment it reads to the end of the cartridge. It also does a few other things:
* Somewhat refactors the SPI flash autodetection code
* Removes jedecid-and-sreg.bin
* Lays groundwork for support of other save types (i. e. WarioWare DIY "Retail NAND")
* [x] Add the save type (and if SPI, the JEDEC ID) to the card info file
* [x] Implement reading the save size from the ExHeader

I still want to do the following in the near future:
* Implement card2 writing based on Corgi3DS
* Figure out the exact boundaries of the writable area on a card2 game
* Attempt to brick a card2 game


d0k3:
Changes look good to me. I'd personally would have handled the card2 save reading in a different (by providing a new function and not modifying `ReadCartSectors`), but it's still a sane and plausible change.

Should I merge the first pull request while you're still working on this one?

dratini0:
Sure, if you don't want to wait much more for wwylele. There is going to be a merge conflict between these two merge requests either way.

On a side note, the good news is that I have [discovered two new commands](https://github.com/dratini0/GodMode9/tree/card2-experiments) on the 3ds cartridge. The bad news is that I am fumbling around in the dark. Do you know what the different values of the `latency` parameter do on `CTR_SendCommand`? And how they interact with the value of `pageSize`? (the commands I seem to have stumbled upon return different data depending on what I set these values to)

d0k3:
I'll just quote @profi200 from IRC here, something to consider:

| 19:29 profi200 https://github.com/dok3/GodMode9/pull/672 Not a good idea sending random cmds.
| 19:30 profi200 Remember Datel Power Saves? (I like to call it Power Bricks)
| 19:31 profi 200 It destroyed CARD2 games in rare cases for unknown reason.
| 19:32 profi200 The cmds to write are known now but it requires some changes to allow writing instead of reading.

dratini0:
Oh, fun... Honestly, I was willing to risk my copy of Pokemon X to find out something that hasn't been done before.

One thing I somewhat expected is that the boundaries of the writable area wouldn't line up with the boundaries of the actual save - I can see that causing bricks. If so, those would be recoverable. But then again, I don't know if that is anywhere near true. (Maybe there is a chip erase command that erases the non-writable region too. That would be terrible)

On the other hand, does this mean that I shouldn't attempt to implement CARD2 save writing for the risk of bricking user's cartridges?

@profi200 - Oooh, care to share that source? I would love to mess with it, even if it will never be merged. The best information I found on the cartridge commands is 3dbrew, which is remarkably scant on information about [command set](http://3dbrew.org/wiki/Gamecards). (GBATEK has roughly the same data, curiously)

profi200:
Pretty much all you need to know (except timeout params) is in [this commit](https://github.com/PSI-Rockin/Corgi3DS/commit/3ecb798c31f185fde4c4cc53b85e6375fc4dfb4d). Again be careful. We don't know what caused these bricks. It may be timing or bad communication between the Power Saves device and card.

dratini0:
Thank you! Another piece of really cool software I knew nothing about! And yes, that should be enough for me to figure it out. I can hopefully guess the timeout parameters, and if not, I can always try running NATIVE_FIRM in Corgi3DS, now that I know it exists.

As for bricking, now I kinda want to replicate the brick, to see what might have caused it, and if it's recoverable (and if my theories hold any water). Though on a second thought, I think I should find something cheaper to deliberately attempt to brick than a Pokemon game...

As for Corgi3DS, I am wondering if it would make a good test harness for gm9? Though it would require the usual dumps to run, so running it in CI might be difficult. I will have to play around with that later.

profi200:
Corgi3DS works mostly but it's in a pretty buggy state. And slow it is. Even with an i7 you are not getting 1 fps. FIRMs not using the GPU or heavy graphics are usable.

dratini0:
Okay, slight update. I have ordered a cheap Card2 game (Animal Crossing: Happy Home Designer) and I will resume card2 experiments once that arrives, using the information from Corgi3DS. This should be later this week or early next week. In the meantime I will implement save size reading.

@profi200 Oh, boy, what a sight is Corgi3DS. I have managed to get to home menu, and even into checkpoint, but it froze after its splash screen (IDK if this had to do with me booting without Luma). On the other hand, every bit of homebew .firm I threw at it fell over one way or another. Most notably, Luma got stuck, I am think waiting for an I2C device to do something.

Honestly, just the fact that I could boot to the home menu using just a bunch of memory dumps is impressive in itself. On the other hand, I find the lack of a GDB stub quite odd. It really makes me wonder what the author's intent was. Either way, I am glad it exists.

profi200:
Corgi is pretty buggy as said but this is to be expected for an emulator still in development. gdb surely is on PSI's list but but not the highest priority.

d0k3:
@dratini0 - just some quick info: I intend to make a new release very soon. There will be a prerelease in the coming days, intended for testing, and a new main release a while after that.

You don't necessarily need to finish this pr until then, if not it will go into the next release. If you want it, in the next release, though, there's still some time to finish it.

dratini0:
Sure thing. Card2 writing isn't done, but at least the sacrificial game is here now. I wouldn't count on that being done "very soon" and reading is fully functional already.

As a footnote, I am pretty sure it will merge conflict with the other one, because I was a bit silly - I recall you wanting me to rebase my branch in a similar situation. Should I do that in this case again?

d0k3:
Merged with current master, thank you! As for that otehr pull request: Yes, pull request based on current master are always preferred, if possible.

PR #674, opened 2021-02-15
ci.yml: changes

Margen67:


aspargas2:
One problem with this: because the devkitarm docker image comes with an older version of git (or at least I think that's why), checkout v2 can't properly clone the repo and instead does it with github's rest API, which leaves out the entire .git folder. This in turn prevents gm9's Makefile from having any information on what commit is being built, so the resulting build artifact will be missing everything derived from that (this can be seen in the artifact produced by the CI build run on this pull request). 

Other than that, these changes are fine. I would recommend simply leaving it using checkout v1, as I do no know of any downside to doing so.

.github/workflows/ci.yml
```diff
@@ -5,10 +5,10 @@ on: [push, pull_request]
 jobs:
   build:
     runs-on: ubuntu-18.04
-    container: devkitpro/devkitarm:latest
+    container: devkitpro/devkitarm
 
     steps:
-      - uses: actions/checkout@v1
```
d0k3, line 11: Leave this at checkout v1 - this will make things easier for us. See @aspargas2's comment.

d0k3 (changes requested):
One small change needed, then we can merge this.

Margen67:
> One problem with this: because the devkitarm docker image comes with an older version of git (or at least I think that's why), checkout v2 can't properly clone the repo and instead does it with github's rest API, which leaves out the entire .git folder. This in turn prevents gm9's Makefile from having any information on what commit is being built, so the resulting build artifact will be missing everything derived from that (this can be seen in the artifact produced by the CI build run on this pull request).
> 
> Other than that, these changes are fine. I would recommend simply leaving it using checkout v1, as I do no know of any downside to doing so.

@aspargas2 Checkout v2 does a [shallow clone by default](https://github.com/actions/checkout#fetch-all-history-for-all-tags-and-branches):
```
# Number of commits to fetch. 0 indicates all history for all branches and tags.
# Default: 1
fetch-depth: ''
```

Margen67:
> 
> 
> One small change needed, then we can merge this.

@d0k3 Fixed.

d0k3 (approved):
Looks good to me!

PR #704, opened 2021-04-22
Loading from certs.db more accordingly + extra Certificate functions

luigoalma:
Instead of guessing offsets, we can load more accordingly by reading the DIFF image and parsing the partitionA data.
This ensures we can find cert, given its imported on the db and the partitionA data is sane.
Even in the cases of imported custom cert names, or dev and retail imported certs mixed in, like in a dev console installing homebrew with the retail certs on cia installs.
Before merge, this could be possibly optimized to not look every time for certs and only once, at least for the default dev and retail certs, but right now, its put up for testing and ensuring it works generally.
A bit of testing appears to show it being functional, however further testing by anyone would be always welcomed.

### DIFF
--- a/arm9/source/game/cert.c
+++ b/arm9/source/game/cert.c
@@ -1,5 +1,6 @@
 #include "cert.h"
 #include "disadiff.h"
+#include "rsa.h"
 
 typedef struct {
     char magic[4]; // "CERT"
@@ -228,6 +229,73 @@ u32 Certificate_GetFullSize(const Certificate* cert, u32* size) {
     *size = sig_size + data_size;
 
     return 0;
+}
+
+static inline u32 _Certificate_KeytypeSignatureSize(u32 keytype) {
+    if (keytype == 0)
+        return 0x200;
+    else if (keytype == 1)
+        return 0x100;
+    else if (keytype == 2)
+        return 0x3C;
+    return 0;
+}
+
+static inline u32 _Certificate_VerifyRSA4096(const Certificate* cert, const void* sig, const void* data, u32 data_size, bool sha256) {
+    (void)cert; (void)sig; (void)data; (void)data_size; (void)sha256;
+    return 2; // not implemented
+}
+
+// noipa, to avoid form of inlining, cloning, etc, to avoid the extra stack usage when unneeded
+static __attribute__((noipa)) bool _Certificate_SetKey2048Misaligned(const Certificate* cert) {
+    u32 mod[2048/8];
+    u32 exp;
+
+    memcpy(mod, cert->data->pub_key_data, 2048/8);
+    exp = getle32(&cert->data->pub_key_data[2048/8]);
+
+    return RSA_setKey2048(3, mod, exp);
+}
+
+static inline u32 _Certificate_VerifyRSA2048(const Certificate* cert, const void* sig, const void* data, u32 data_size, bool sha256) {
+    if (!sha256)
+        return 2; // not implemented
+
+    int ret;
+
+    if (((u32)&cert->data->pub_key_data[0]) & 0x3)
+        ret = !_Certificate_SetKey2048Misaligned(cert);
+    else
+        ret = !RSA_setKey2048(3, (const u32*)(const void*)&cert->data->pub_key_data[0], getle32(&cert->data->pub_key_data[2048/8]));
+
+    if (ret)
+        return ret;
+
+    return !RSA_verify2048(sig, data, data_size);
+}
+
+static inline u32 _Certificate_VerifyECC(const Certificate* cert, const void* sig, const void* data, u32 data_size, bool sha256) {
+    (void)cert; (void)sig; (void)data; (void)data_size; (void)sha256;
+    return 2; // not implemented
+}
+
+u32 Certificate_VerifySignatureBlock(const Certificate* cert, const void* sig, u32 sig_size, const void* data, u32 data_size, bool sha256) {
+    if (!sig || !sig_size || (!data && data_size) || !Certificate_IsValid(cert))
+        return 1;
+
+    u32 keytype = getbe32(cert->data->keytype);
+    u32 _sig_size = _Certificate_KeytypeSignatureSize(keytype);
+
+    if (sig_size != _sig_size)
+        return 1;
+
+    if (keytype == 0)
+        return _Certificate_VerifyRSA4096(cert, sig, data, data_size, sha256);
+    if (keytype == 1)
+        return _Certificate_VerifyRSA2048(cert, sig, data, data_size, sha256);
+    if (keytype == 2)
+        return _Certificate_VerifyECC(cert, sig, data, data_size, sha256);
+    return 1;
 }
 
 static u32 _Certificate_AllocCopyOutImpl(const Certificate* cert, Certificate* out_cert) {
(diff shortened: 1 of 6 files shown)


arm9/source/game/cert.c
```diff
@@ -1,21 +1,319 @@
...
-u32 LoadCertFromCertDb(u64 offset, Certificate* cert, u32* mod, u32* exp) {
-    Certificate cert_local;
-    FIL db;
-    UINT bytes_read;
+typedef struct {
+    char magic[4]; // "CERT"
+    u8 unk[4]; // afaik, always 0
+    u8 used_size[4]; // size used after this header
+    u8 garbage[4]; // literally garbage values
+} PACKED_STRUCT Certsdb_partition_header;
```
d0k3, line 9: Better (and matching the current style): CertsDbPartitionHeader

arm9/source/game/cert.c
```diff
@@ -1,21 +1,319 @@
...
+
+        if (snprintf(full_issuer, 0x41, "%s-%s", cert_local.data->issuer, cert_local.data->name) > 0x40)
+            break;
+
+        if (!strcmp(full_issuer, issuer)) {
+            ret = 0;
+            break;
+        }
+
+        free(cert_local.sig);
```
d0k3, line 300: Shouldn't you use the cleanup function you wrote here?

arm9/source/game/cert.c
```diff
@@ -1,21 +1,319 @@
...
+        free(cert_local.data);
+        cert_local.sig = NULL;
+        cert_local.data = NULL;
+
+        offset += full_size;
+    }
+
+    if (ret) {
+        free(cert_local.sig);
```
d0k3, line 309: Same here, it's cleaner to use the function you wrote.

arm9/source/game/cia.c
```diff
@@ -59,21 +60,49 @@ u32 BuildCiaCert(u8* ciacert) {
         0x18, 0x83, 0xAF, 0xE0, 0xF4, 0xE5, 0x62, 0xBA, 0x69, 0xEE, 0x72, 0x2A, 0xC2, 0x4E, 0x95, 0xB3
     };
 
-    // open certs.db file on SysNAND
-    FIL db;
-    UINT bytes_read;
-    if (f_open(&db, "1:/dbs/certs.db", FA_READ | FA_OPEN_EXISTING) != FR_OK)
+    const char* issuer_ca = !IS_DEVKIT ? "Root-CA00000003" : "Root-CA00000004";
```
d0k3, line 63: This applies to `ticket.c` and `tmd.c` as well: I see a lot of repeated code here. Wouldn't it be possible to do all the repeating stuff in a function in certs.c? You shouldn't even have a lot of parameters here.

arm9/source/game/ticket.c
```diff
@@ -28,17 +28,26 @@ u32 ValidateTwlTicket(Ticket* ticket) {
...
-    static bool got_modexp = false;
-    static u32 mod[0x100 / 0x4] = { 0 };
-    static u32 exp = 0;
-
-    if (!got_modexp) {
-        // grab mod/exp from cert from cert.db
-        if (LoadCertFromCertDb(0x3F10, NULL, mod, &exp) == 0)
-            got_modexp = true;
-        else return 1;
+    Certificate cert;
```
d0k3, line 31: Take note that the previous version of this function stored the mod and exp, so it wouldn't have to go through `certs.db` every time. Performance is important here, so we should keep the mod and exp once we got it. Same applies to `tmd.c

d0k3 (changes requested):
Alright, thanks for this pull request. I've only got a few minor complaints, some stuff that should be changed before the merge. other than that, looking very good!

d0k3:
Also, sorry about the radio silence. Real life and stuff :^).

luigoalma:
No problems on the silent! Alright lets see about some of these change requests

arm9/source/game/cert.c
luigoalma: Fixed in commit 809d031

arm9/source/game/cert.c
luigoalma: Fixed with commit d346524

arm9/source/game/cert.c
luigoalma: Also fixed with commit d346524

arm9/source/game/ticket.c
luigoalma: Indeed, however as i pointed out on my initial PR comment, only still up for general testing functionality, in regards to mod and exp saving, reasons i removed it for now is that sometimes, a console may have retail and devkit tickets, tmds and or certificates mixed in. And I wonder if i should provide a static storage for the common retail/dev certificates, and fetch from that storage on cert.c if requested again after their first load.

arm9/source/game/cia.c
luigoalma: It can be, but given that `cia.c` also loads the same way, but 3 certs, I could make a function that takes a variable argument list (`...`), or a list of certificate names in a `const char**` list terminated by a NULL or a count argument, that then generate a raw certificate bundle.

arm9/source/game/ticket.c
d0k3: Yeah, didn't think about that. Storing mod and exp is important here, though. Maybe store the type (uninitialized, dev, retail) of stored mod and exp here, and only reload if it changes?

luigoalma: I can do a static storage with slots for expected common entries, and give them when they were already loaded before on a `Certificate` object, and make Cleanup call check if it is from static storage to not call free on the wrong addresses.

luigoalma: Commit aec3989 brings less loading by storing certs whenever they load. My `ticket.db` took about 14 seconds to sort tickets before, now takes about 4 seconds. But still can improve loading modulus with extra functions for `Certificate`.

arm9/source/game/cia.c
luigoalma: Commit 5b27b5a, asides a bit of refactoring, implements a cleaner call, and more optimized, as there's no repeated calls of `LoadCertFromCertDb` as effect.

luigoalma:
There's still something I'd like to implement before merge.
Saying in case change requests and reviews pass before I do.

arm9/source/game/ticket.c
luigoalma: Commit 78c2bc4 adds signature verification functions against certificate, avoiding also copying out modulus, at least when unneeded.

luigoalma:
I believe I done all the edits I can think of, now its just waiting review, and extra code change requests if any. Also testing by anyone interested in testing it too, even if tests seem to pass correctly on my side, extra testing is welcome.

d0k3:
Alright, after another review, this looks good to me. Would you be okay with squashing this to fewer commits before the merge? Sorry about being a bit slow these days, too.

So, @luigoalma, is this ready to be merged?

luigoalma:
It's now squashed to 7 commits, it's about as much as I can safely rewrite history and squash them, unless you'd like them to be squashed to 1 commit. And yes, at least by me, it's ready to merge.

d0k3 (approved):

d0k3:
Merged with current master. Thanks for your contribution!

PR #710, opened 2021-06-04
Detect transparent pixels in DS icon and set them to white

WaluigiWare64:
This PR adds transparency detecting code to `GetTwlIcon()` and sets the transparent pixels to white (like in the 3DS Home Screen).
Example with _Dr. Kawashima's Brain Training: How Old Is Your Brain?_
**Before:**
snap_210604112826
**After:**
snap_210604112843

Here is an image of how it looks on the 3DS Home Screen:
2021-06-04_12-32-19 444_bot


d0k3:
Wouldn't it be better to set that pixel to black? (same as the GodMode9 background color)

WaluigiWare64:
> 
> 
> Wouldn't it be better to set that pixel to black? (same as the GodMode9 background color)

@d0k3 we could, but in my opinion it would be better to show icons how they are displayed on the 3DS (and DSi) home screen. On the DS, it is transparent, but behind it is a white/grey gradient.
I don't know any icons like this, but some icons may not look good on a black background with (fake) transparency.

aspargas2:
For what it's worth, I'm inclined to agree with setting them to white for consistency with the home menu.

Wolfvak:
@WaluigiWare64 in the interest of consistency, can you preserve the palette[0] value and restore it before returning? Thanks!

d0k3:
Alright ,sorry, I know this took long. I'm requesting one small change...

arm9/source/game/nds.c
```diff
@@ -133,7 +133,9 @@ u32 GetTwlTitle(char* desc, const TwlIconData* twl_icon) {
 u32 GetTwlIcon(u16* icon, const TwlIconData* twl_icon) {
     const u32 h = TWLICON_DIM_ICON; // fixed size
     const u32 w = TWLICON_DIM_ICON; // fixed size
-    const u16* palette = twl_icon->palette;
+    u16* palette = twl_icon->palette;
+    u16 palette_first = palette[0];
+    palette[0] = 0x7FFF;
```
d0k3, line 138: You shouldn't write here, even if you set the value back later on, there's a better solution....

d0k3 (changes requested):
In general, this is a good change, but the change I requested will still improve on this.

arm9/source/game/nds.c
WaluigiWare64: I did that before, but @Wolfvak recommended to do this instead:
```
<Wolfvak> you can restore the original value after you're done?
<Wolfvak> it's much faster than branching
```

Wolfvak: Yeah, I messed up and didn't notice it was `const`. If you wanna keep this approach you'd have to copy the palette to a local buffer.
I'm not sure if it's worth it, when comparing the setup time of this temp buffer to the time needed by the conversion time itself.

WaluigiWare64: @d0k3 I reverted that change and went back to my original method, which is the way you suggested.

d0k3:
Merged to master. Thanks for your contribution! (and sorry this, again, took so long)

PR #717, opened 2021-07-05
Fix NDS / DSi names being cut off when renaming to good name

TimmSkiller:
To better understand what this fixes, let's take a look at what happens when renaming this NDS file to good name:
og_prompt
og_result

The result is a cut-off name, because currently, the name is copied *only until the first newline*.
NDS/DSi Names consist of `Title + Publisher` in one string. However, to avoid overlapping, parts of the Title and the Publisher are separated by newlines, as seen here:
grafik

and this is what the Title looks like, if we were to interpret it as a string:
```c
"Professor Kageyama's\nMaths Training\nNINTENDO"
```
Therefore, __every character until the last newline is still part of the Title, because what comes after the last newline is the Publisher__.

I have implemented it like this:
1. Instead of checking for the **first** occurrence of a newline, I'm checking for the **last** ocurrence of a newline, to know where the Title stops and the Publisher starts.
2. Looping through the Title and replacing every remaining newline with a space to merge all possible parts together, forming the correct Title.

And as a result:
fix_prompt
fix_result
It now correctly renames the Title, it is no longer cut off.

### DIFF
--- a/arm9/source/utils/gameutil.c
+++ b/arm9/source/utils/gameutil.c
@@ -4014,8 +4014,15 @@ u32 GetGoodName(char* name, const char* path, bool quick) {
         } else {
             char title_name[0x80+1] = { 0 };
             if (GetTwlTitle(title_name, (TwlIconData*) icon) != 0) return 1;
-            char* linebrk = strchr(title_name, '\n');
+            char* linebrk = strrchr(title_name, '\n'); // search for the last occurence of newline, because everything until the last newline is part of the name
+
             if (linebrk) *linebrk = '\0';
+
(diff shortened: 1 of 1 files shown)


TimmSkiller:
After some more testing, it seems that it works fine with DSiWare CIAs, too.
before:
`00030015534C524E TWiLight Menu++ (TWL-SLRN) (DSi Exclusive) (W) (v0.0.0).standard.cia`
after:
`00030015534C524E TWiLight Menu++ Last-run ROM (TWL-SLRN) (DSi Exclusive) (W) (v0.0.0).standard.cia`

d0k3:
Thanks, good addition!

PR #722, opened 2021-07-16
Preserve RSA keys in trimmed NTR dumps

### LINKED ISSUE #721: Fix exclusion of RSA key in trimmed NTR image
With the release of nds-bootstrap v0.44.2, the issue it was having with cloneboot (a means of Download Play) was fixed. The issue was that an RSA key was not being read. This RSA key is at the end of the game data, so an extra 0x88 bytes needs to be read from NTR cartridges. Reading this extra data was implemented into GM9i v2.7.0.

Relevant code snippet (difference at the end):
Before:
`if (trimRom) {
	romSize = (isDSiMode() && (ndsCardHeader.unitCode != 0) && (ndsCardHeader.twlRomSize > 0))
		? ndsCardHeader.twlRomSize : ndsCardHeader.romSize;
}`
After:
`if (trimRom) {
	romSize = (isDSiMode() && (ndsCardHeader.unitCode != 0) && (ndsCardHeader.twlRomSize > 0))
		? ndsCardHeader.twlRomSize : ndsCardHeader.romSize + 0x88;
}`

lifehackerhansol:
Fixes #721, implemented exactly as described in the issue.

This was tested on a Mario Kart DS (Korea) cartridge, while being played on nds-bootstrap and an original R4, where cloneboot requires the RSA key. It is useful for real hardware NDS playback as a whole.

I unfortunately do not have a TWL cart to test the fix there. If anyone else is willing to test, or have any suggestions about what to add to this PR, feel free to let me know or directly commit.

### DIFF
--- a/arm9/source/gamecart/gamecart.c
+++ b/arm9/source/gamecart/gamecart.c
@@ -246,7 +246,7 @@ u32 InitCartRead(CartData* cdata) {
         if (nds_header->device_capacity >= 15) return 1; // too big, not valid
         if (cdata->cart_size == 0)
             cdata->cart_size = (128 * 1024) << nds_header->device_capacity;
-        cdata->data_size = nds_header->ntr_rom_size;
+        cdata->data_size = nds_header->ntr_rom_size + 0x88;
         cdata->arm9i_rom_offset = 0;
 
         // TWL header
--- a/arm9/source/gamecart/gamecart.c
+++ b/arm9/source/gamecart/gamecart.c
@@ -246,8 +246,6 @@ u32 InitCartRead(CartData* cdata) {
         if (nds_header->device_capacity >= 15) return 1; // too big, not valid
         if (cdata->cart_size == 0)
             cdata->cart_size = (128 * 1024) << nds_header->device_capacity;
-		// safety check for 0x88 addition
-		if (cdata->data_size > nds_header->ntr_rom_size + 0x88) return 1;
         cdata->data_size = nds_header->ntr_rom_size + 0x88;
         cdata->arm9i_rom_offset = 0;
 
--- a/arm9/source/gamecart/gamecart.c
+++ b/arm9/source/gamecart/gamecart.c
@@ -246,6 +246,8 @@ u32 InitCartRead(CartData* cdata) {
         if (nds_header->device_capacity >= 15) return 1; // too big, not valid
         if (cdata->cart_size == 0)
             cdata->cart_size = (128 * 1024) << nds_header->device_capacity;
+		// safety check for 0x88 addition
+		if (cdata->data_size > nds_header->ntr_rom_size + 0x88) return 1;
         cdata->data_size = nds_header->ntr_rom_size + 0x88;
         cdata->arm9i_rom_offset = 0;
 


lifehackerhansol:
Force push was to change the author email. Nothing else changed.

lifehackerhansol:
There have been some points brought up mid-testing with some members in the community as to whether it is actually worthwhile to keep the appended 0x88 in TWL carts. There isn't a known TWL cart (as in DSi Enhanced or DSi Exclusive) that utilizes cloneboot at all, so it may just be unneeded extra data which goes against the purpose of trim dumps.

In GodMode9i, where this was first implemented (mentioned in the related issue), I don't think there is a particular check for NTR or TWL dumps. In my opinion if it'll be done on one cart it should be done for all, but I think it's better for someone else to decide this.

SNBeast:
There is a check for NTR vs. TWL dumps in GodMode9i, and it's in the code snippet in the linked issue above: `ndsCardHeader.unitCode != 0`. The unit code is 0 for NTR, 2 for TWL with NTR support, and 3 for TWL only.

lifehackerhansol:
I see, so the check only applies when the unit code is 0, then? I may have misunderstood the code the first time I saw.

SNBeast:
Yes. `+` has a higher operator precedence than `?` and `:`, so the choices read "`ndsCardHeader.twlRomSize` if true, `ndsCardHeader.romSize + 0x88` if false."

lifehackerhansol:
Yeah not sure how I messed that up. It's gone now.

lifehackerhansol:
At the moment, this commit freezes GodMode9, though it is implemented as described in the issue. I'll try again.

d0k3:
Should not freeze GM9. Make sure you've got the latest master as base. Also take note that appending + 0x88 to that exact location is what I did earlier. According to @SNBeast this is not correct.

lifehackerhansol:
> Should not freeze GM9. Make sure you've got the latest master as base. Also take note that appending + 0x88 to that exact location is what I did earlier. According to @SNBeast this is not correct.

Apparently, it doesn't work when reading from C: GAMECART. It does however, preserve the key on an existing ROM on the SD.

d0k3:
Alright, I forgot about that one actually, that's a different place in the code. Maybe your initial fix wasn't so bad. You need to make sure, tough, that the data size never exceeds the cart size.

d0k3:
Don't want to bother, but do you by any chance have a new build that can be tested? Sorry about the earlier confusion, your initial fix was actually the correct way, it just needs an additional sanity check build in. If you need more info, just ask.

lifehackerhansol:
I haven't had time to do more commits. It's been on my mind for quite a while though, I might revert the changes again. 

I have absolutely no idea where I'd place the checks, though.

d0k3:
It belongs right below where you made the data_size change. data_size + 32 may, in some rare cases, exceed cart_size, causing a lot of trouble. You just make sure data_size is never more than cart_size.

lifehackerhansol:
Isn't that check already done in https://github.com/d0k3/GodMode9/blob/4dc96d37e84e341cf8e0aecc1a26c1e0f84b4004/arm9/source/gamecart/gamecart.c#L274 ? Or do you want more safety checks?

d0k3 (changes requested):
That check is wrong, but, guess what, you can remove it. We already got the sanity check, a few lines below:
`if (cdata->data_size > cdata->cart_size) return 1;` (which you already noticed, I see).

Anyways, remove the check. And yes, I'm fully aware we're then back at the first version of this pull request. And yes, my fault for not paying attention and causing all the confusion. Sorry :^)

Once you done that, could you provide a testing built in the GitHub issue? I'll merge it if it is tested okay then.

lifehackerhansol:
Kind of off-topic, but I noticed local builds seem to produce smaller binaries... but I assume the end result would be the same, so.

d0k3:
Merged, thank you!

PR #728, opened 2021-09-01
Allow loading FRF fonts up to 0x20000 bytes and add Misaki Gothic 8x8 font to resources

Epicpkmn11:
- 0x10000 bytes is a bit too small for complete coverage of JIS X 0201 and probably similar for a reasonably complete Chinese font
   - I think 0x20000 should be fine for any single language at least and it doesn't seem to cause any issues for me, sorry if this breaks anything I'm not noticing 😅 
- Adds 美咲ゴシック (Misaki Gothic) 8x8 Japanese font to the resources folder
   - Didn't want to do a full Japanese font with Kanji for the default since it's quite large, I think it's probably good to have in the resources folder though
   - This font is 0x11308 bytes, so just a bit over the old limit
   - It has complete coverage of [JIS X 0201](https://en.wikipedia.org/wiki/JIS_X_0201), which I think is basically everything you could want for Japanese and also partial Cyrillic and Greek (Greek is missing the Tonos letters and I think Cyrillic is missing a couple too)
   - I also edited it slightly to have a copy of the full width latin and katakana for ASCII and half-width katakana as while the font does have proper half-width letters since GodMode9 is monospace it looks better with the full-width ones (the first two rows in the PBM)
   - There is also a 6x8 version, I went with the 8x8 version though as while the Latin is rather spaced out the Kanji is more readable and if you're using this font it's probably because you want kanji
   - License is here: https://littlelimit.net/font.htm#license
- If someone knows some good Korean or Chinese bitmap fonts that are the right size I could do those too, I don't know enough of those languages to find one easily

<details><summary>Images</summary>

Font: (blank areas are ignored as I have them all mapped to duplicates of space or full-width space in the mapping file)
misaki_gothic

Screenshot:
snap_210901151101

</details>


d0k3:
Merged to the `unicode` branch. 0x20000 should be perfectly okay, even that's just 8kiB, so no worries. Thanks for the ultra-quick response and your work in this!

PR #736, opened 2021-09-16
Make TWL MBR check actually give useful info

fox8091:
Currently, if TWL MBR is corrupt, the error shows "Error: MBR MBR is corrupt", while CTR MBR being incorrect shows "Error: CTR MBR is corrupt". This changes this to match the error style given for CTR MBR errors.


fox8091:
Don't actually merge this, a better solution is being worked on that overhauls the NAND validation function to allow checking for multiple problems, as well as to fix a few other bugs that affect devkits.

PR #737, opened 2021-09-16
Add a new type of flash chip, from Art Academy

dratini0:
Yesterday evening, Discord user FerozElMejor has discovered a new kind of save chip in a copy of Art Academy. With @Epicpkmn11's help, we have surmised its JEDEC ID is `0xC22017`, and based on that it's manufactured by Macronix. We have tested the write settings as in the commit, and they seem to work too.

The one other notable change is changing the timeout on the program operation. This seemed necessary for FerozElMajor's cartridge.


d0k3:
Merged, thanks a lot!

PR #740, opened 2021-09-26
Improved markdown viewer feature

pokemaniac05:
I added markdown viewer (referred as `MD ToC Viewer`) feature, as seen in `HOME` -> `More...` -> `Show ReadME` to every `.md` file in sdcard and other partitions. To view these file, user has to press A and select `View in MD ToC Viewer`. I also added a `Markdown...` menu to view files inside `0:/gm9/markdown` (like `Scripts...` and `Payloads...` menus).
Finally I edited `Show ReadME` to enter .md file name via soft keyboard (to let user view other .md files added to VRAM at compile time).

I tested code changes on Old2DS and they work.


d0k3:
There are several issues with this, the most important is: That ToC viewer is not designed for universal support (it does even say so in the source code comments). So, with any given .md file, anything could happen - even if all the files you tested caused no trouble. You also used tabs instead of spaces (GodMode9 uses spaces) and the MD support directory plus the additional HOME menu entry is too much for something that's not a "main" feature (this is still GodMode9, not MarkDown9 now).

That being said, I'm not against a universal Markdown viewer, but it also needs universal compatibility then, and it would need no more integration but the A button menu. Can't merge this like it is, so I'm closing it for now.

pokemaniac05:
I resolved the bug reported by @aspargas2 (now the buffer is initialized with VRAM0_README_MD, as everyone might except) and added security check if given .md file has no captions (that file will be opened by TextViewer instead ToC). I think that @d0k3 was referring to that.
Regarding indentation, I noticed that a lot of code (also files that I have not touched) uses, for example, 8 char tabs and two spaces instead of 10 spaces and I have not changed that aspect. 
May @d0k3 or @aspargas2 reopen PR?

PR #741, opened 2021-09-30
Add user-facing SHA1 support

aspargas2:
This PR adds an option in the GUI to calculate the SHA-1 hash of any file, and adds a `-1` flag to the `sha`, `shaget`, and `cp -h` scripting commands to use SHA-1 instead of the default SHA-256. It also allows for characters `'0'` through `'5'` to be used as flags for scripting commands, and fixes a bug in the scripting engine where `err_str` would not be set if an invalid character was passed as a flag.

### DIFF
--- a/arm9/source/filesys/fsutil.c
+++ b/arm9/source/filesys/fsutil.c
@@ -161,7 +161,7 @@ size_t FileGetSize(const char* path) {
     return fno.fsize;
 }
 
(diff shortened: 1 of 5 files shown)


d0k3:
Merged to master. I only slightly changed the display of SHA-1 checksums via an additional commit.

PR #745, opened 2021-10-10
Improved markdown viewer (fixed #740)

pokemaniac05:
This PR is a fixed version of #740 
I resolved the bug reported by @aspargas2 (now the buffer is initialized with VRAM0_README_MD, as everyone might except) and added security check if given .md file has no captions (that file will be opened by TextViewer instead ToC). I think that @d0k3 was referring to that.
Regarding indentation, I noticed that a lot of code (also files that I have not touched) uses, for example, 8 char tabs and two spaces instead of 10 spaces and I have not changed that aspect.

### DIFF
--- a/arm9/source/utils/scripting.c
+++ b/arm9/source/utils/scripting.c
@@ -1730,6 +1730,7 @@ bool MemTextViewer(const char* text, u32 len, u32 start, bool as_script) {
 
 // right now really only intended for use with the GodMode9 readme
 // (misses safety checks for wider compatibility)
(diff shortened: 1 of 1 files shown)


d0k3:
What @Wolfvak wrote - we can't merge this to GodMode9 because the included markdown viewer does not have the compatibility or was ever intended to be a universal markdown viewer. It just works with the readme and provides a convenient way to navigate it, but that's that. Even though I'll have to close this, know that contributions are always appreciated.

Issue #758, opened 2022-01-21
[FEATURE REQUEST] Add creation date to File Info


blueberryfigapricotrye:
**Got a great idea on how to improve GodMode9?**
Currently GodMode9 v2.1.0 shows only "Modified" timestamps at File Info. For me it is too little details. 
The "Created" timestamps should be displayed.
![snap_220121024057](https://user-images.githubusercontent.com/98134989/150450240-47c85b14-f2bf-43a2-ac68-4a9d7981681e.png)

**Describe the feature you'd like**
I'm used to seeing "Modified" and "Created" timestamps in the default file manager on my Windows and macOS notebook when I use my SD card from the 3DS or local external hard drive. Here what I mean exactly.

File Explorer (Windows 11)
![1](https://user-images.githubusercontent.com/98134989/150450725-9a3dfba7-8bee-4825-8a76-242a5939ebc7.PNG)

Finder (macOS 12)
![2](https://user-images.githubusercontent.com/98134989/150450421-40899e50-10d8-40d6-8f2b-87b35d0a9354.PNG)

**Describe alternatives you've considered**
There is no alternative.

novialriptide:
I'm reading the source code right now and it seems like the `created` timestamp is already implemented in its code. Are you sure GodMode9 doesn't have an existing option to see the `created` timestamp? I don't have my 2DS XL onhand right now to check.

novialriptide:
I'm looking through `FILINFO` [documentation ](http://elm-chan.org/fsw/ff/doc/sfileinfo.html) right now and it seems like this isn't possible. If someone could correct me, that would be great.

TimmSkiller:
It shows `modified` for files and `created` for directories. This is because `FILINFO` contains the "last modified" date in case of files and the "created" date/time in case of directories:
> fdate: The date when the file was modified or the directory was created.
> ftime: The time when the file was modified or the directory was created.

So yeah, not possible.

Wolfvak:
Technically speaking the FAT Filesystem itself doesn't even support creation timestamps, only last modification.
There are semi-standard extensions that allow this by storing the data in a reserved area but they're not supported by FATFS, and even if they were they're basically a hack (for Win9x) and we have no way to guarantee it'll work on all FAT-capable machines.

Closing this as not possible, as said by the two messages above.

blueberryfigapricotrye:
Thank you for the quick responses. I have some unexplained questions now.



[...]

PR #759, opened 2022-01-28
changes to add created/modified/last accessed date

ZeroSkill1:
**I know this isn't exactly a good idea.**

As mentioned in issue #758, GodMode9 currently does not show file creation date. It shows whatever FatFs returns in `FILINFO`, which is creation date (in case of directories) and modification date (in case of files). 

According to [this document](https://cscie92.dce.harvard.edu/spring2021/slides/FAT32%20File%20Structure.pdf), it's possible to get:
- Creation Date + Time
- Modification Date + Time
- Last Accessed Date (no Time here)

This may not seem like a good idea at first, but I got it working by modifying FatFs a tiny bit.
I made sure not to mess with anything that isn't related. I also made sure the dates are correct for drives and virtual drives, keeping the date that was previously intended to be shown.

`FILINFO` only stored one Date and Time, so the first thing I did was adding new fields to store the other Dates/Times.
Then, I made sure the Dates/Times were updated accordingly (for writes, file/dir creation, etc.).

This is what I got after that:
grafik

The data shown is identical to the data shown in the Windows Explorer "Properties" menu, so it is at the very least correct.
I have also tested the modified date, by creating a file, modifying it in the built in hex editor and saving it. The modified date/last access date is updated accordingly.

Overall it's a pretty minimal change. I understand that it may not be something you wish to add, but if it works, IMO, why not.

Let me know what you think.

### DIFF
--- a/arm9/source/fatfs/ff.c
+++ b/arm9/source/fatfs/ff.c
@@ -149,9 +149,11 @@
 #define DIR_NTres			12		/* Lower case flag (BYTE) */
 #define DIR_CrtTime10		13		/* Created time sub-second (BYTE) */
 #define DIR_CrtTime			14		/* Created time (DWORD) */
+#define DIR_CrtDate         16      /* Created date (WORD) */
 #define DIR_LstAccDate		18		/* Last accessed date (WORD) */
 #define DIR_FstClusHI		20		/* Higher 16-bit of first cluster (WORD) */
 #define DIR_ModTime			22		/* Modified time (DWORD) */
+#define DIR_ModDate         24      /* Modified date (WORD) */
 #define DIR_FstClusLO		26		/* Lower 16-bit of first cluster (WORD) */
 #define DIR_FileSize		28		/* File size (DWORD) */
 #define LDIR_Ord			0		/* LFN: LFN order and LLE flag (BYTE) */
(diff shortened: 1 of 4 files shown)


arm9/source/fatfs/ff.c
```diff
@@ -3766,9 +3771,12 @@ FRESULT f_open (
 				} else
 #endif
 				{
+					DWORD tm = GET_FATTIME();
 					/* Set directory entry initial state */
 					cl = ld_clust(fs, dj.dir);			/* Get current cluster chain */
-					st_dword(dj.dir + DIR_CrtTime, GET_FATTIME());	/* Set created time */
+					st_dword(dj.dir + DIR_CrtTime, tm);	/* Set created time */
+					st_dword(dj.dir + DIR_ModTime, tm);	/* Set modified time */
+					st_word(dj.dir + DIR_LstAccDate, (WORD)(((tm) >> 16) & 0xFFFF)); /* Set last access date */
```
eku, line 3779: Convert `(WORD)(((tm) >> 16) & 0xFFFF)` to a macro, as it is used several times.

eku (commented):
Please check your settings for code indentation and spaces versus tabs.

d0k3:
@Wolfvak - I'm still unsure about this one. While I like the functionality, and think @ZeroSkill1 did a good implementation here, it will make FATFS updating hard. However an update may not be required, as [most recent updates in FATFS](http://elm-chan.org/fsw/ff/updates.html) deals with ExFAT, which we're not really interested in. @aspargas2 - what do you think?

PR #761, opened 2022-02-22
Improve DSiWare save generation

Epicpkmn11:
The sector/track/head count calculation for DSiWare saves was a bit off resulting in some DSiWare refusing to boot because the save 'didn't have enough room' or such. If you want to test, Dragon Quest Wars (EUR) is one game that was broken, theoretically any game with a 32 KiB save (or like half the other sizes) should be broken but most probably don't bother checking.

I've tested this calculation against all 25 sizes that TWiLight Menu++ has (it has [hardcoded template files](https://github.com/DS-Homebrew/TWiLightMenu/tree/b3ca0f376e499545ada022db4757cd8c407085fe/romsel_dsimenutheme/nitrofiles/DSiWareSaveHeaders) at the moment) and everything seems to generate correctly with this change.

### DIFF
--- a/arm9/source/game/nds.c
+++ b/arm9/source/game/nds.c
@@ -55,20 +55,24 @@ u32 BuildTwlSaveHeader(void* sav, u32 size) {
     u16 n_sct = 1;
     u16 sct_track = 1;
     u16 sct_heads = 1;
-    while (true) {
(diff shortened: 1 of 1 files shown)


d0k3:
Merged, thanks a lot for this!

PR #778, opened 2022-06-23
use sizeof with snprintf when target is an array

gemarcano:
- GCC warned of a case when the size specified in a snprintf call was
   larger than the size of a target buffer. To fix, when possible use
   sizeof() to match buffer size.

I tried to only modify snprintf calls that worked on buffers with known sizes, as sometimes they're used with buffers passed in as parameters (with expected sizes, but can't be determined via `sizeof`). Hopefully none of these snuck though.

I haven't tested this yet on my 3DS. I'll update the state to not a draft once I can confirm I can boot this.

### DIFF
--- a/arm9/source/utils/gameutil.c
+++ b/arm9/source/utils/gameutil.c
@@ -313,7 +313,7 @@ u32 LoadTicketForTitleId(Ticket** ticket, const u64 title_id) {
     // path to ticket.db
     char path_ticketdb[32];
     char drv = *path_store;
-    snprintf(path_ticketdb, 32, "%2.2s/dbs/ticket.db",
+    snprintf(path_ticketdb, sizeof(path_ticketdb), "%2.2s/dbs/ticket.db",
         ((drv == 'B') || (drv == '5') || (drv == '4')) ? "4:" : "1:");
 
     // load ticket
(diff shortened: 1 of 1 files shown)


gemarcano:
Just checked, G9 boots fine on my 3DS. Confirmed that is a build I made as of today and not an old copy.

arm9/source/utils/gameutil.c
```diff
@@ -3560,27 +3560,27 @@ u32 ShowGameCheckerInfo(const char* path) {
...
         (type & GAME_CDNTMD) ? "CDN" :
         (type & GAME_TWLTMD) ? "TWL" : "UNK");
 
     char contents_str[64];
-    if (type & GAME_CIA) snprintf(contents_str, 64, "Contents in CIA: %lu/%lu", content_found, content_count);
+    if (type & GAME_CIA) snprintf(contents_str, sizeof(contents_str), "Contents in CIA: %lu/%lu", content_found, content_count);
     else snprintf(contents_str, 64, "Contents in TMD: %lu", content_count);
 
     char conid_str[32] = { '\0' };
-    if (type & (GAME_CIA|GAME_TIE)) snprintf(conid_str, 64, "Console ID: %08lX\n", console_id);
```
Wolfvak, line 3583: good catch

PR #803, opened 2023-03-22
Fix buffer overflow when encrypting/decrypting CIA files non-inplace

ZeroSkill1:
The issue would occur when encrypting or decrypting CIA files (**that have a meta region**) to `0:/gm9/out`. It is caused by passing a pointer to a buffer that is not large enough to hold the CIA meta section. The solution is to allocate a separate buffer, as has been done in other places where the meta section is copied.


d0k3:
Approved and taken over. Thank you!

PR #807, opened 2023-04-15
Use the correct MCU LED period for calculations

### LINKED ISSUE #772: [BUG] Info LED code making the LED pulse 32 times faster than intended
3dbrew and by extension GodMode9 are wrong about the notification LED data format and units

The color data is `u8[32]` per color component, each index representing a "frame". The MCU runs the periodic task of updating that LED at a 512Hz rate.

- `delay` is the delay/period in "task" (512Hz) ticks between two array elements, not counting the current task tick. Which means it's `clamp((512 * delayMs / 1000) - 1, 1, 255)`. 0 will cause an underflow. This limits the period to 500ms exactly.
- `smoothing` is the fade-out time. Same unit and range as `delay` (1..255). You usually put the same value as `delay` (smooth transition), or 1 (blink).
- `loop_time` is the amount of time to wait before restarting the pattern, *not* counting the current tick. It can be 0 (restart immediately), or 255 ("don't restart")

See here: https://github.com/SciresM/boot9strap/blob/master/stage2/arm9/source/utils.c#L89

Hope this helps.

Wolfvak:
Better late than never

Fixes #772

### DIFF
--- a/arm11/source/hw/mcu.c
+++ b/arm11/source/hw/mcu.c
@@ -22,6 +22,8 @@
 
 #include <stdatomic.h>
 
+#include "common.h"
+
 #include "arm/timer.h"
 
 #include "hw/gpio.h"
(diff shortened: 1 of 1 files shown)


d0k3 (approved):
Looks good to me, feel free to merge!

PR #808, opened 2023-04-16
Fix compilation warnings on dkA r60

Wolfvak:
Mostly just switched out the bound strn* functions for their unbound variants when the string is a known constant


d0k3 (approved):
Looks good to me!

PR #809, opened 2023-04-16
Refactor to improve modularity

Nemris:
This PR adds documentation and type annotations, and allows the script to be imported as a module.

Additionally, the input file is closed as soon as the data is read and the output file is written
to in a single go, while the TRF data structures are populated in memory.

### DIFF
--- a/utils/transriff.py
+++ b/utils/transriff.py
@@ -1,66 +1,300 @@
 #!/usr/bin/env python3
 
-from argparse import ArgumentParser
-from os import path
+""" Create an TRF translation for GodMode9 from a translation JSON. """
 
+from __future__ import annotations
+
+import argparse
+import dataclasses
 import json
+import math
+import pathlib
 import struct
+import sys
 
-parser = ArgumentParser(description="Creates an TRF translation for GodMode9 from a translation JSON")
-parser.add_argument("input", type=str, help="JSON to convert from")
-parser.add_argument("output", type=str, help="to output to")
-parser.add_argument("version", type=int, help="translation version, from language.inl")
 
-args = parser.parse_args()
+LANGUAGE_NAME = "GM9_LANGUAGE"
 
-with open(args.input, "r") as f:
-    # read JSON
-    strings = json.load(f)
-    if "GM9_LANGUAGE" not in strings:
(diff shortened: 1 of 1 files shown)


Epicpkmn11:
This is my fault but everywhere that says "an TRF" should actually be "a TRF", I copied this from fontriff.py and it seems forgot to fix that whoops, might be a good idea to refactor fontriff.py too if this style is preferred

---

Personal opinion, feel free to ignore: imo this is... not exactly great? like, it takes 5 times as much code to say the same thing which makes it far less readable to me while also breaking Python 2 compatibility... I'd definitely be for moving the main code to a function called in an `if __name__ == "__main__":` (as is my more recent style, fontriff.py was a while ago), but I can't say I'm personally a fan of this style.

Nemris:
I can technically just scrap the rewrite and focus on deviating minimally for the original, if preferred. Off the top of my head, the three changes I'd make for sure in that case are to remove the unused import, add the usual `main()` and limit the scope of the handle to the JSON file.
I can also add documentation if needed, but the type hints might have to be skipped if Python 2 compatibility is wanted.

Epicpkmn11:
I'd rewrite them like this personally: https://gist.github.com/Epicpkmn11/6869a7862e83c76a630337a1ccc44758

Not necessarily saying that's better, I guess I just don't much like OOP in Python

Wolfvak:
I think I already mentioned some of these points in other chats but I'll list them here anyways:

- Yay for static typing
- There's _a ton_ of boilerplate code for converting between format and others, is there no way to skip it or work in a single format?
- Documentation is good, but this seems a bit excessive...
- +291 lines, -57 lines changed, I know this is Python code (and I'm no expert by a long shot) but it seems like there might be some unnecessary complexity?
- Don't care for Python2 compatibility, most distros and systems don't even have installers or packages for it

Nemris:
Sounds good - should I get started on applying the recommendations or wait for @aspargas2?

d0k3:
Sorry this took me so long. @Epicpkmn11 - do you think this is ready to merge?

Epicpkmn11:
Yeah I think this should be good

d0k3:
Merged, thank you

PR #816, opened 2023-06-16
Correct installable address whitelist

danny8376:
blacklist is stored at 0xb088 - 0xb0bf, which is ulong[14] in {start(inclusive), end(exclusive)} pair.
one thing to note is that boot9 use inclusive comparing with blacklist start for both section load address and
section load address + section size (comparing logic is at 0xa42e - 0xa449), so if the firm fits perfectly at the end of the space right before the blacklisted range, it'll also be rejected.


danny8376:
This is mainly by-product when I'm trying to make stupidly huge vram drive for use with ntrboot.
I test it by installing firm directly as flashing ntrboot is relatively annoying. Then when i installed a firm that's supposed to exactly full occupy arm9 memory, it triggered bootrom error for some reason.
After digging into boot9, figured that it does not only blacklist vector table, but also blacklist the last 32KiB of arm9 memory (not sure what it's actually used for)

danny8376:
also, find out it's also documented in gbatek (well, kinda expected)
https://problemkaputt.de/gbatek-3ds-files-firm-format.htm

arm9/source/system/vram0.h
```diff
@@ -18,7 +18,7 @@
 #define VRAM0_EASTER_BIN       "easter.bin"
 
 
-#define VRAM0_OFFSET    0x080C0000
+#define VRAM0_OFFSET    0x080B7E00
```
Wolfvak, line 21: What's up with needing this?

Wolfvak (commented):
> Can you give us some background information, maybe some reference for that change? Also, @Wolfvak - does this need to be fixed?

With nocash's reference, it looks good to me. The one thing that's kind of bugging me is the initrd/vram drive load address change, is there any reason for that?

If we're really starved for space in there, we might as well fully switch over to making it an object bundled in the ARM9 binary rather than a separate section as it is now - it was done that way because originally it was kept in VRAM (which is why it's called that way), but since we made the change to keeping it in internal memory I'd say we go all the way and make it an ARM9 binary section (not FIRM section)

arm9/source/system/vram0.h
danny8376: as commit said, to avoid bleeding into blacklisted range
or we can just shrink tar size limit 0x8200 bytes more to 223.5 KiB
I just prefer pretty-ish size over address plus it's already pretty small
(and we still have plenty space for actual arm9 binary)

Wolfvak: At that point I think we might as well just go with my other suggestion and embed the tar as an object in the ARM9 binary, rather than a section in FIRM.

0x080C0000 is already a pretty magical number, but 0x080B7E00 is even more so, and I'd rather avoid these.

danny8376: Well, then i guess just shrink the size even more for now, ppl need more will try to hack it themselves i think

Wolfvak (approved):

PR #820, opened 2023-07-06
Add minimal hold time for cancel of ShowProgress

danny8376:
This is mainly for scripting, for fixing not ideal hold b to cancel behavior
```
fdummy -o -s 9:/t1 2000000
ask -o -s "WAH"
cp 9:/t1 9:/t2
# on fast drive, press b on the ask above will also let cp catch the button press
# also, if you press b to cancel the cancel, it'll also register again,
# as the copying process is too fast that the delay between two button checks is way too short
# (it's checked every 1MiB transfer currently)

fdummy -o -s 0:/test 100
ask -o -s "WAH"
cp 0:/test 1:/test
# similar thing will also happen when cp very small file,
# as it's done very quickly, then it check button press
# altough, it's technically done, the prompt will still appear
```

### DIFF
--- a/arm9/source/common/ui.c
+++ b/arm9/source/common/ui.c
@@ -21,6 +21,7 @@
 #define FONT_MAX_WIDTH 8
 #define FONT_MAX_HEIGHT 10
 #define PROGRESS_REFRESH_RATE 30 // the progress bar is only allowed to draw to screen every X milliseconds
+#define PROGRESS_CANCEL_MIN_HOLD_MS 1000
 
 typedef struct {
     char chunk_id[4]; // NOT null terminated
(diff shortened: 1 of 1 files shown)


d0k3:
Not so sure about that one. The way it's done just seems overly complicated, and we don't actually have a fast drive. Wouldn't it be better to make sure the B keypress is not registered multiple times?

danny8376:
> Not so sure about that one. The way it's done just seems overly complicated, and we don't actually have a fast drive. Wouldn't it be better to make sure the B keypress is not registered multiple times?

well, the 2nd case of my example will likely trigger this issue with literally any sd card i believe
assuming you unlock write permission beforehand, which indeed the issue we have during implement new 3ds.hacks.guide finalizing script
yeah, 0x100 bytes might be too extreme, but a few KiB file will surely trigger this constantly
I mean we encounter this when we're copying GM9Megascript from sd to sd, it won't trigger on every sd card but do trigger on 2 of 3 total testers when we consider we're actually using slow sd card

PR #824, opened 2023-07-22
Embed the VRAM tar data in the ARM9 executable

Wolfvak:
As a follow up to some of my comments on #816, this PR removes the extra FIRM section that was used as the "VRAM drive" (huge misnomer for a while but whatever, my bad) and instead embeds it as part of the ARM9 section, compiled as part of the `arm9.elf` binary.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,7 +89,9 @@ $(VRAM_TAR): $(SPLASH) $(OVERRIDE_FONT) $(VRAM_DATA) $(VRAM_SCRIPTS)
 	@echo "Building $@"
 	@$(MAKE) --no-print-directory -C $(@D)
 
-firm: $(ELF) $(VRAM_TAR)
+arm9/arm9.elf: $(VRAM_TAR)
+
+firm: $(ELF)
 	@mkdir -p $(call dirname,"$(FIRM)") $(call dirname,"$(FIRMD)")
 	@echo "[FLAVOR] $(FLAVOR)"
 	@echo "[VERSION] $(VERSION)"
--- a/Makefile
+++ b/Makefile
@@ -16,9 +16,9 @@ export RELDIR := release
 export COMMON_DIR := ../common
 
 # Definitions for initial RAM disk
-VRAM_OUT    := $(OUTDIR)/vram0.tar
+VRAM_TAR    := $(OUTDIR)/vram0.tar
 VRAM_DATA   := data
-VRAM_FLAGS  := --make-new --path-limit 99 --size-limit 228864
+VRAM_FLAGS  := --make-new --path-limit 99 --size-limit 262144
 ifeq ($(NTRBOOT),1)
 	VRAM_SCRIPTS := resources/gm9/scripts
 endif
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -80,24 +80,25 @@ release: clean unmarked_readme
...
 	@-7za a $(RELDIR)/$(FLAVOR)-$(VERSION)-$(DBUILTS).zip ./$(RELDIR)/*
 
-vram0:
-	@mkdir -p "$(OUTDIR)"
-	@echo "Creating $(VRAM_OUT)"
-	@$(PY3) utils/add2tar.py $(VRAM_FLAGS) $(VRAM_OUT) $(shell ls -d $(SPLASH) $(OVERRIDE_FONT) $(VRAM_DATA)/* $(VRAM_SCRIPTS))
+$(VRAM_TAR): $(SPLASH) $(OVERRIDE_FONT) $(VRAM_DATA) $(VRAM_SCRIPTS)
+	@mkdir -p "$(@D)"
+	@echo "Creating $@"
+	@$(PY3) utils/add2tar.py $(VRAM_FLAGS) $(VRAM_TAR) $(shell find $^ -type f)
```
Wolfvak, line 86: Kind of unnecessary changes but now there's no real need to force a rebuild unless the source files change, I guess it's more "correct"

Makefile
```diff
@@ -80,24 +80,25 @@ release: clean unmarked_readme
...
 
 %.elf: .FORCE
 	@echo "Building $@"
 	@$(MAKE) --no-print-directory -C $(@D)
 
-firm: $(ELF) vram0
-	@test `wc -c <$(VRAM_OUT)` -le 228864
```
Wolfvak, line 93: No need to test for limits, the compiler will throw an error if we exceed the ARM9 section limits

arm9/source/system/vram_data.s
```diff
@@ -0,0 +1,12 @@
+.section .rodata.vram_data
+
+.align 2
+.global vram_data
+vram_data:
+	.incbin "../output/vram0.tar"
```
Wolfvak, line 6: I'm not thrilled with the hardcoded path, or even worse making a reference to something a directory up, but it works and needed minimal changes to implement.

Makefile
```diff
@@ -80,24 +80,27 @@ release: clean unmarked_readme
...
+arm9/arm9.elf: $(VRAM_TAR)
```
Wolfvak, line 92: Technically the ARM9 now depends on the .tar file itself, so it must be built before the ARM9 binary. They could be done in parallel, but the difference is so small it's preferable to be safe (in case the user doesn't have Python installed or something)

Wolfvak:
@danny8376 any comments on this? iirc you were testing the limits of the VRAM drive sizes, this might help you get larger drives bundled in GM9

danny8376:
fine with it? I do have two different cases for large vram drive
one is the normal-ish one that can fit within arm9 mem.
for that, this will surely make it easier as there's no need to tweak offset manually anymore
(but i'm actually fine with both way as tweaking the offset isn't that annoying)
another one is well over 1MiB, like full 6MiB VRAM or even more
but that's totally abnormal and hacky and need way more changes, so not really related here
also, the change isn't that huge, so doesn't affect that much

aspargas2 (approved):
Looks good to me :+1:

PR #827, opened 2023-08-20
Move the start of AHBWRAM to be immediately after the VECTORS

### LINKED ISSUE #826: [BUG] AHBWRAM section can overflow into hardcoded BootROM copy space
Main issue is there's [b9s bootrom dump](https://github.com/d0k3/GodMode9/blob/031762a1fe7af334a1b9ff5225c5f8e3927fa977/arm9/source/system/memmap.h#L41-L45) which we can't move
so [AHBWRAM](https://github.com/d0k3/GodMode9/blob/master/arm9/link.ld#L7) should actually be shrunken to 488K to avoid overlapping with boot9 bootrom dump (0x08080000 - 0x08006000 = 488K)
Since we only have 488K, minus 377K (arm9.bin of v2.1.1), we actually only have 111K available, which is less than half of what we have previously (223.5K)

And do we actually have anything located at 0x080A0000 ~ 0x080C0000? if not, we should be able to shift old vram0 to 0x080A0000 and enlarge it to 351.5K, or we can simply ignore bootrom dump and enlarge AHBWRAM directly?

Wolfvak:
Fixes #826


d0k3 (approved):
Looks good to me, but you know much better than me anyways :).

PR #828, opened 2023-08-26
arm9/arm11 Makefile: Fix building with spaces in path

ihaveamac:
This quotes $(CURDIR) which fixes an issue with building GM9 in a path with spaces, which Windows users are likely to have.


Wolfvak (approved):
classic mistake, thanks!

PR #837, opened 2024-01-20
Split GM9 across two sections in memory.

DarkRTA:
This is needed to implement support for Lua. Lua is way to big to fit in a single contiguous section of memory without overwriting our copy of boot9 and boot11, so some clever linker manipulation was used to place things in the free space after it. LTO also had to be disabled as we won't be able to split `.text` later with it.

This uses the following memory map:

```
AHBWRAM: 0x08000040 - 0x08080000
    .vectors
    .text
    .rodata
    .data
    .bss
    .rodata
    .data
    .bss

AHBWRAM2: 0x080a0000 - 0x08100000
    .text (files in arm9/source/lua)
    .text (libm)
    .rodata.vram_data

```

This has been tested on actual hardware and seems to boot up and work fine.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,10 +106,10 @@ firm: $(ELF) $(BINS)
 	@echo "[BUILD] $(DBUILTL)"
 	@echo "[FIRM] $(FIRM)"
 	@$(PY3) -m firmtool build $(FIRM) $(FTFLAGS) -g -D $(BINS) arm11/arm11.elf \
-		-A 0x08000000 0x080a0000 -C NDMA NDMA XDMA -n 0x08000040
+		-A 0x08000040 0x080a0000 -C NDMA NDMA XDMA -n 0x08000080
 	@echo "[FIRM] $(FIRMD)"
 	@$(PY3) -m firmtool build $(FIRMD) $(FTDFLAGS) -g -D $(BINS) arm11/arm11.elf \
-		-A 0x08000000 0x080a0000 -C NDMA NDMA XDMA -n 0x08000040
+		-A 0x08000040 0x080a0000 -C NDMA NDMA XDMA -n 0x08000080
 
 vram0: $(VRAM_TAR) .FORCE # legacy target name
 
--- a/Makefile
+++ b/Makefile
@@ -91,15 +91,25 @@ $(VRAM_TAR): $(SPLASH) $(OVERRIDE_FONT) $(VRAM_DATA) $(VRAM_SCRIPTS)
 
 arm9/arm9.elf: $(VRAM_TAR)
 
-firm: $(ELF)
+$(OUTDIR)/AHBWRAM.bin: $(ELF)
+	@$(OBJCOPY) -O binary arm9/arm9.elf --only-section=AHBWRAM $@
+
+$(OUTDIR)/AHBWRAM2.bin: $(ELF)
+	@$(OBJCOPY) -O binary arm9/arm9.elf --only-section=AHBWRAM2 $@
+
+BINS := $(OUTDIR)/AHBWRAM.bin $(OUTDIR)/AHBWRAM2.bin
+
+firm: $(ELF) $(BINS)
 	@mkdir -p $(call dirname,"$(FIRM)") $(call dirname,"$(FIRMD)")
 	@echo "[FLAVOR] $(FLAVOR)"
 	@echo "[VERSION] $(VERSION)"
 	@echo "[BUILD] $(DBUILTL)"
 	@echo "[FIRM] $(FIRM)"
-	@$(PY3) -m firmtool build $(FIRM) $(FTFLAGS) -g -D $(ELF) -C NDMA XDMA
+	@$(PY3) -m firmtool build $(FIRM) $(FTFLAGS) -g -D $(BINS) arm11/arm11.elf \
+		-A 0x08000000 0x080a0000 -C NDMA NDMA XDMA -n 0x08000040
 	@echo "[FIRM] $(FIRMD)"
-	@$(PY3) -m firmtool build $(FIRMD) $(FTDFLAGS) -g -D $(ELF) -C NDMA XDMA
+	@$(PY3) -m firmtool build $(FIRMD) $(FTDFLAGS) -g -D $(BINS) arm11/arm11.elf \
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -91,15 +91,25 @@ $(VRAM_TAR): $(SPLASH) $(OVERRIDE_FONT) $(VRAM_DATA) $(VRAM_SCRIPTS)
...
 	@echo "[VERSION] $(VERSION)"
 	@echo "[BUILD] $(DBUILTL)"
 	@echo "[FIRM] $(FIRM)"
-	@$(PY3) -m firmtool build $(FIRM) $(FTFLAGS) -g -D $(ELF) -C NDMA XDMA
+	@$(PY3) -m firmtool build $(FIRM) $(FTFLAGS) -g -D $(BINS) arm11/arm11.elf \
+		-A 0x08000000 0x080a0000 -C NDMA NDMA XDMA -n 0x08000040
 	@echo "[FIRM] $(FIRMD)"
-	@$(PY3) -m firmtool build $(FIRMD) $(FTDFLAGS) -g -D $(ELF) -C NDMA XDMA
+	@$(PY3) -m firmtool build $(FIRMD) $(FTDFLAGS) -g -D $(BINS) arm11/arm11.elf \
+		-A 0x08000000 0x080a0000 -C NDMA NDMA XDMA -n 0x08000040
```
DarkRTA, line 112: just out of curiosity, is there way you'd like for this to actually be done?

Makefile
```diff
@@ -91,15 +91,25 @@ $(VRAM_TAR): $(SPLASH) $(OVERRIDE_FONT) $(VRAM_DATA) $(VRAM_SCRIPTS)
...
 arm9/arm9.elf: $(VRAM_TAR)
 
-firm: $(ELF)
+$(OUTDIR)/AHBWRAM.bin: $(ELF)
+	@$(OBJCOPY) -O binary arm9/arm9.elf --only-section=AHBWRAM $@
+
+$(OUTDIR)/AHBWRAM2.bin: $(ELF)
+	@$(OBJCOPY) -O binary arm9/arm9.elf --only-section=AHBWRAM2 $@
+
+BINS := $(OUTDIR)/AHBWRAM.bin $(OUTDIR)/AHBWRAM2.bin
```
Wolfvak, line 100: Do these need to be .bin files? Objcopy should be able to extract only the section we care about while preserving LMA/VMA information as an .elf

arm9/Makefile
```diff
@@ -10,8 +10,8 @@ INCDIRS := source source/common source/filesys source/crypto source/fatfs source
 INCLUDE := $(foreach dir,$(INCDIRS),-I"$(shell pwd)/$(dir)")
 
 ASFLAGS += $(SUBARCH) $(INCLUDE)
-CFLAGS  += $(SUBARCH) $(INCLUDE) -fno-builtin-memcpy -flto
```
Wolfvak, line 13: Not sure why we have to disable LTO, what error do you get when compiling?

arm9/link.ld
```diff
@@ -4,44 +4,54 @@ ENTRY(_start)
 
 MEMORY
 {
-    VECTORS (RX) : ORIGIN = 0x08000000, LENGTH = 64
     AHBWRAM (RWX) : ORIGIN = 0x08000040, LENGTH = 512K - 64
+    AHBWRAM2 (RWX) : ORIGIN = 0x080a0000, LENGTH = 384K
```
Wolfvak, line 8: Can you rename to `AHBWRAM_LO` and `AHBWRAM_HI`? Or `_1` and `_2`?

arm9/link.ld
```diff
@@ -4,44 +4,54 @@ ENTRY(_start)
...
+
+
+    __vectors_lma = 0x08000000;
+    AHBWRAM : ALIGN(4) {
+        /* this needs to be absolute otherwise GM9 won't boot */
+        __vectors_vma = ABSOLUTE(.);
+        KEEP(*(.vectors));
+        . = ALIGN(4);
+        __vectors_len = . - __vectors_vma;
+        . = . + 0x10;
```
Wolfvak, line 41: Why is this extra space required? Just curious

arm9/Makefile
DarkRTA: Okay this is.mainly for more control over the linking.

With LTO enabled it becomes impossible to move some functions to a different translation unit.

arm9/link.ld
DarkRTA: This space was present at the end of what used to be `.vectors`.

Makefile
DarkRTA: Seems to also work fine as an `.elf` file but it may take be a bit to figure out why objcopy is throwing a warning.

Makefile
```diff
@@ -91,15 +91,25 @@ $(VRAM_TAR): $(SPLASH) $(OVERRIDE_FONT) $(VRAM_DATA) $(VRAM_SCRIPTS)
 
 arm9/arm9.elf: $(VRAM_TAR)
 
-firm: $(ELF)
+$(OUTDIR)/AHBWRAM_LO.elf: $(ELF)
+	@$(OBJCOPY) arm9/arm9.elf -j AHBWRAM_LO $@
+
+$(OUTDIR)/AHBWRAM_HI.elf: $(ELF)
+	@$(OBJCOPY) arm9/arm9.elf -j AHBWRAM_HI $@
```
DarkRTA, line 98: `objcopy` is now throwing a warning and it does not provide any option to silence it.
Since this doesn't break the build, do you have any suggestions on how I can silence its output.

ihaveamac:
Trying some stuff with this PR and I think a few features aren't working properly. Specifically I tried building a CIA for a built-in title (Activity Log USA 0004001000021200), and it froze. Building the latest commit of the main repo (ad8b5e0a8c7c9d7dd590fb76a29fe5d6ab18e22f) and it works fine.

Is there a chance something else is using arm9 memory? Or is something just weird with my setup? I don't know what else could have broken.

ihaveamac:
Maybe related to lto? Because building the latest mainline commit without lto results in a crash:
IMG_0238

DarkRTA:
I'd open a separate issue for that. This reeks of UB.

DarkRTA:
Wolfvak has offered a much better solution than what is present here. Closing this.

Issue #841, opened 2024-03-04
[BUG] No NAND access - New 2DS XL Panda


Lethaltail:
Issue: No access to SysNAND or EmuNAND on assumedly all development units. Only S: and E: virtual drives will be available alongside the SD Card and 9/C/M/V:. System info will show <unknown> in 9/13 fields. 
Bisect shows commit 203cf7f to be when dev NAND access broke. Only tested on a JAN panda.

Steps to reproduce: 
- Build 203cf7f or newer for dev (or grab 1.9.3pre1 or later)
- Boot your build on dev
- Observe the issue

Screenshots:
![image](https://github.com/d0k3/GodMode9/assets/13040464/3df98783-05f5-4136-8c1d-824d721a565d)

![image](https://github.com/d0k3/GodMode9/assets/13040464/052e7e00-4d6d-45ab-b8ec-633ee2ad0e4f)

Console Type: JAN-DEV (Model: New 2DS XL (panda))
No hardware modifications 
Bootloader: boot9strap
Loading Method: GodMode9 as boot.firm (Can also Luma chainload for same effect)

**Have you actually read the template?**
[X] I have read the information

urherenow:
I think you can build yourself, replacing the aeskeydb.bin with the legit (non-obfuscated) one, and your panda should work.

Wolfvak:
fixed in https://github.com/d0k3/GodMode9/commit/26990ca23a4011428ec93fd4b24ab1f0b47acc0f

PR #842, opened 2024-03-11
Fix decryption of keys in aeskeydb.bin for dev

luigoalma:
Change Key Y to generate same Normal key on developer consoles with 0x2C
Reference to #841


Lethaltail:
Tested and working on the same system as #841.

PR #845, opened 2024-03-17
Update Makefile

Ricca665:
Fixes python3 not found when running make on windows


d0k3:
This works for me when compiling in Windows.

PR #846, opened 2024-03-17
Create build.bat

Ricca665:


Wolfvak:
unnecessary
this is probably slower than just running `make`, which is all that's needed

PR #847, opened 2024-03-17
Create build.sh

Ricca665:


urherenow:
I would be surprised to see this accepted in the first place, but even if it would be, I’m sure the maintainers would recommend a proper shebang.

#!/bin/sh

because 1) it’s just a script, and 2) not everyone uses bash as their default shell.

Wolfvak:
same as #846

PR #865, opened 2024-06-15
Use VA start and end linker symbols instead of length

### LINKED ISSUE #859: GodMode9 compiled with gcc 14.1.0 hangs at boot
As discussed on the server here is the issue.

How to reproduce:
Compile GodMode9 with devkitARM r64 (gcc 14.1.0) and try to boot it from Luma3DS or fastboot3DS. It will hang on blackscreens (no backlight).
Have not tried to compile under Windows so can't say for sure if the behavior differs but i doubt it. It's likely undefined behavior somewhere that gets detected and optimized in a way that breaks the code.

Probably unrelated to the actual issue this new warning popped up:
```
lto-wrapper: warning: using serial compilation of 8 LTRANS jobs
lto-wrapper: note: see the '-flto' option documentation for more information
/opt/devkitpro/devkitARM/bin/../lib/gcc/arm-none-eabi/14.1.0/../../../../arm-none-eabi/bin/ld: warning: arm9.elf has a LOAD segment with RWX permissions
```

Wolfvak:
Fixes #859

### DIFF
--- a/arm11/link.ld
+++ b/arm11/link.ld
@@ -16,7 +16,7 @@ SECTIONS
         __text_va = ABSOLUTE(.);
         *(.text*)
         . = ALIGN(4K);
-        __text_len = . - __text_va;
+        __text_va_end = .;
     } >AXIWRAM
 
     .data : ALIGN(4K)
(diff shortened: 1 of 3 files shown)


profi200:
Works fine now.

d0k3 (approved):
Can't tell you guys much about that ARM stuff, you know much better than me. The code looks good, I'm okay with merging it.

PR #877, opened 2024-11-08
Fix multi RedNAND setup in SD format menu

ZeroSkill1:
The SD format menu allows the setup of 2x, 3x, or 4x multi RedNANDs.

Using the 3x and 4x options works properly, but using the 2x option, the SysNAND is only copied to the space reserved for the first RedNAND, leaving the other one without proper setup.

This results in the inability to access this EmuNAND space even when attempting to use the `Switch EmuNAND` option in the `more...` menu.

I have provided a fix which should allow for proper intialization of multi RedNAND setups, while keeping support for single Gateway type EmuNANDs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,7 @@ jobs:
           rm $ZIPNAME
           echo "OUTNAME=${ZIPNAME%.zip}" >> $GITHUB_ENV
 
-      - uses: actions/upload-artifact@v2
+      - uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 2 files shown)


d0k3:
Looks good to me and can be merged. But, why does the automatic check fail? It's not about the changes in this pr.

ZeroSkill1:
I updated the workflow so it succeeds now.

Test build can be downloaded [here](https://github.com/d0k3/GodMode9/actions/runs/11741609104/artifacts/2162788129).

d0k3:
Merged, thank you!

PR #881, opened 2024-11-30
New translations

Epicpkmn11:
- Updates translation files
   - New translation and 100% complete: Chinese (simplified)
   - German is also up to 100% complete now
- Adds a new font for Chinese and Korean
   - No Korean translation as of yet, but I prepped the font since the font I used has it
   - The font is from https://github.com/TakWolf/fusion-pixel-font, with the common half-width characters swapped out for the old 8x8 GM9 font as mixed width fonts are unsupported
- Removed Misaki Gothic files from the repo as they are redundant

I would suggest merging #880 first as that will probably create conflicts on the TRF files, hence opening this as a draft.


Wolfvak (approved):

d0k3:
Merged, thank you!

PR #883, opened 2024-12-01
Split up the ARM9 code and data sections

Wolfvak:
Split up the ARM9 code (.text, .vectors) and data (.rodata, .data, .bss) sections into their own ELFs. This allows us to use more ARM9 WRAM while leaving the 128k BootROM mirror intact.

It also removes the warning about having an RWX segment, though it also creates warnings about having empty segments (`arm-none-eabi-objcopy: arm9.elf: warning: empty loadable segment detected at vaddr=0x80a0000, is this intentional?`, still looking into the exact flags to get rid of it).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 *.obj
 *.elf
 *.map
+*.dis
 
 # Precompiled Headers
 *.gch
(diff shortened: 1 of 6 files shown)


d0k3:
Merged, thanks!

PR #884, opened 2024-12-17
Add sdmount and sdumount scripting commands

MisterSheeple:
These are basically just switchsd but without forcing the user to switch SD cards. Made this so that I can reload the emunand easily.


ihaveamac:
Just wondering, does nextemu not do this already? Even with a single emunand?

MisterSheeple:
Oh, thanks. That isn't in any documentation so I didn't know it was a thing, but it's exactly what I need. I'll leave this pr open in case this would be a useful feature for anything else, but that solves it for me.

PR #894, opened 2025-03-21
Fix Chinese font loading

danny8376:
Fix buffer allocation properly so Chinese fonts can be loaded properly
(without this fix, it can only be loaded when switching language since it's the only place that allocate buffer by font size)


Epicpkmn11 (approved):
This looks good, I just stuck with the arbitrary buffers because that's how it was before, but I kept running into issues with the CJK ones cause well... Those fonts get large.

Wolfvak (approved):

d0k3:
Merged, thank you!

PR #895, opened 2025-03-21
Tweaks for Chinese font

danny8376:
replace ascii characters (0x0021-0x007E) with full-width characters (0xFF01-0xFF5E) so the overall font looks more balanced

Original:
snap_250321203824

Tweaked:
snap_250321214321


danny8376:
I'd assume ja and ko have similar issue and need to apply the same tweak, but nothing to test with (japanese doesn't use this font and currently korean translation doesn't exist) so i decide to not touch them for now.

d0k3:
Merged, thank you!

PR #903, opened 2025-03-25
Fix compiling on non-UTF8 system, especially Windows

R-YaTian:


lkdjfalnnlvz (approved):

Makefile
```diff
@@ -34,11 +34,7 @@ JSON_FILES     := $(filter-out $(SOURCE_JSON),$(wildcard $(JSON_FOLDER)/*.json))
 TRF_FILES      := $(subst $(JSON_FOLDER),$(TRF_FOLDER),$(JSON_FILES:.json=.trf))
 
 ifeq ($(OS),Windows_NT)
-	ifeq ($(TERM),cygwin)
-		PY3 := py -3 # Windows / CMD/PowerShell
-	else
-		PY3 := py # Windows / MSYS2
-	endif
+	PY3 := py -3 # Windows / CMD / PowerShell
```
d0k3, line 37: Why did you make that change?

Makefile
```diff
@@ -34,11 +34,7 @@ JSON_FILES     := $(filter-out $(SOURCE_JSON),$(wildcard $(JSON_FOLDER)/*.json))
 TRF_FILES      := $(subst $(JSON_FOLDER),$(TRF_FOLDER),$(JSON_FILES:.json=.trf))
 
 ifeq ($(OS),Windows_NT)
-	ifeq ($(TERM),cygwin)
```
Wolfvak, line 37: We've already changed this python detection mechanism several times, can you post a source on why this is the correct way to detect it across all platforms?

R-YaTian: > We've already changed this python detection mechanism several times, can you post a source on why this is the correct way to detect it across all platforms?

py -3 is a good choice for windows CMD and/or powershell. But compare $TERM with cygwin is not a correctly way. Normally "$TERM == cygwin" will always false when using CMD/Powershell, even MSYS terminal. Then we will always go to "PY3: py" branches. That's where the problem lies. If developers install both CPython_Win32 and cpython_mingw, "PY3: py" will use cpython_mingw which will cause "Win32 error 6". So we need to use "py -3" for workaround this.

Makefile
R-YaTian: > Why did you make that change?

"py -3" is a good choice for windows CMD and/or powershell. But compare $TERM with cygwin is not a correctly way. Normally "$TERM == cygwin" will always false when using CMD/Powershell, even MSYS terminal. Then we will always go to "PY3: py" branches. That's where the problem lies. If developers install both CPython_Win32 and cpython_mingw, "PY3: py" will use cpython_mingw which will cause "Win32 error 5/6". So we need to use "py -3" for workaround this.

eku:
May I suggest to split this PR since it contains changes for two different problems. Please don't mix.

d0k3:
Still unsure about the py -3 change. We could of course take this over partially (only the UTF-8 fixes). I'm not working with Windows anymore, so I cannot test or even give an educated opinion on this on.

@Wolfvak - what do you think?

Wolfvak:
> @Wolfvak - what do you think?

I think we can partially merge this, excluding the py -3 change.

d0k3:
Alright. @R-YaTian - can you remove that change? Afterwards we'll merge it.

R-YaTian:
> Alright. @R-YaTian - can you remove that change? Afterwards we'll merge it.

Done!

Wolfvak (approved):

d0k3:
.. merged.

PR #904, opened 2025-03-25
fix internalfs_make_dummy_file causing a data abort on failure

ZeroSkill1:
`internalfs_make_dummy_file` would cause a data abort on failure due to a missing variadic function parameter.


ihaveamac:
lgtm

PR #905, opened 2025-03-25
fix a few cases of improper varg usage in `luaL_error`

ZeroSkill1:
There were a few cases of improper/missing variadic function parameters in calls to `luaL_error`, which would cause a data abort / invalid error message output on failure.


ihaveamac:
lgtm

PR #916, opened 2025-04-19
GM9Megascript: remove "Un-install Hax"

ihaveamac:
This option is confusing because despite it saying "will completely remove CFW", it is not sufficient to actually fully remove custom firmware. It doesn't, and can't, detect everything like region changes, modified system titles, and other things that could brick a stock console.

There have been numerous cases of bricked consoles because someone saw this, thinking it would do everything needed to restore a console back to an unmodified state, and getting a bricked console that needs a flashcart purchase to fix.

(There is also a issue where, even if everything else was in order, it can still brick a console due to a bad SD card. For some reason it copies the FIRM to the SD card, where it can potentially get corrupted if a counterfeit card was used.)

A proper uninstall guide is required, and we have one here: https://3ds.hacks.guide/uninstall-cfw


d0k3:
Agreed and merged

PR #920, opened 2025-06-17
Correct the region detection logic

heiyu-04:
Correct the region detection logic


luigoalma:
I understand that, however my opinion still stands on regards to C.
sys could provide instead region as the effective region (from C, or A or B if there's no C) as one variable and another as real region from non C.
To me, main letter of secinfo should still be the real secinfo's letter and then another variable to confirm C as existent.
A change in behavior can affect existing scripts as well, this could also lead to unattended side effects to some others.
So new variables with more information sounds better than changing current ones, this still would allow scripts to change around an addition rather than a change on a main behavior.
Up to someone to review this further and give their opinion too.

PR #938, opened 2025-10-30
GM9 Lua: "draw" class

nika5thgearluffy:
This is a draft PR that adds support for drawing pixels, boxes, and text on Lua scripts.

While Lua scripting has only benefited processing commands on GodMod9, I thought it would be crazy to extend it to have features useful for creating other things such as games and apps.

I may not know the scope of GodMode9's scripting, but I wanted to try something new, so this exists.

So, yep.

--------

Functions added:

- draw.pixel(screenNumber, x, y, color) = Draws a pixel at a specified point at whatever screen you choose.
- draw.box(screenNumber, x, y, width, height, color) = Draws a filled box at a specified point at whatever screen you choose.
- draw.text(screenNumber, text, x, y, color) = Draws text with the font that's being used at a specified point at whatever screen you choose.
- draw.rgb(red, green, blue) = Creates an RGB color that can be used for any of the draw functions. Use this once in a variable for consistency.

- ui.clear_full() = Clears all screens than just the screen that's being used for scripts.

(Will add more if more is added)

### DIFF
--- a/arm9/source/lua/gm9draw.c
+++ b/arm9/source/lua/gm9draw.c
@@ -1,0 +1,93 @@
+#ifndef NO_LUA
+
+#include "gm9draw.h"
+#include "common/ui.h"
+#include "common/colors.h"
+
+static int draw_pixel(lua_State *L) {
+    CheckLuaArgCount(L, 4, "draw.pixel");
+    u16* screen;
+    
+    int screenID = lua_tonumber(L, 1);
+    if (screenID == 1) {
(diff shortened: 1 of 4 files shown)


Gruetzig:
While i think this wont lead to games, I was thinking of a custom UI and for what I have in mind arbitrary rectangle and text drawing would be very appreciated. So, I wanted to ask about the state of this draft PR. If you don't have anything else you want to add to this, could we move forward towards merging this?

PR #947, opened 2026-02-03
Scripting: Normalize output of FormatBytes

Epicpkmn11:
The output of `FormatBytes()` is localized, but it is provided to scripts and some (such as the hacks.guide [finalizing script](https://github.com/hacks-guide/finalize/blob/mane/romfs/finalize/finalize.gm9#L184-L198)) expect it in the English format. Currently trying to run finalize with any locale that uses `,` as the decimal separator crashes the script and anything that renames "GB" (ex. French "Go") would cause that script to incorrectly display the no free space error if those languages didn't all use "," separator.

I considered making the Lua version have a flag to enable/disable localization, but I couldn't quite tell what the proper way to do that would be.

### DIFF
--- a/arm9/source/lua/gm9lua.h
+++ b/arm9/source/lua/gm9lua.h
@@ -14,10 +14,11 @@
 #define EXPLORER        (1UL<<16)
 #define ENCRYPTED       (1UL<<17)
 #define SIG_CHECK       (1UL<<18)
+#define USE_LOCALE      (1UL<<19)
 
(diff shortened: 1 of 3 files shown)


Wolfvak (approved):

Epicpkmn11:
Oh didn't see your second comment so I just wrote it myself :sweat_smile: 

Fixed the stray `;` and updated the docs, did it in a force push cause I know I've been asked to keep my commits clean here before sometimes.

ihaveamac:
It’s fine, yours is better than what I typed up anyway.

PR #953, opened 2026-03-23
Fix install destination detection for CXI-based game files

### LINKED ISSUE #952: [BUG] incorrect nand/sd app detection
---
name: Bug report
about: Incorrect detection of nand/sd app when installing
title: "[BUG] incorrect nand/sd app detection"
labels: bug
assignees: ''

---

Currently, gm9 uses the title ID of an app to determine if a title is going to be (or is) installed on the NAND or SD.
While this works in most cases, some titles (such as devmenu, config, and some others), uses a SD title ID, but is actually a NAND app.

What's resulting is that the app is getting installed to the wrong location, the app is usable when installed to the sd, but isn't where it should be (and can potentially get issues). Also, this makes the uninstallation of such title giving an error because it's expecting to find it on the SD card, but is in the NAND.

To fix this, we have to check the NCCH flag instead of relying on the titleID of the app.

ZeroSkill1:
Should fix #952. GodMode9 currently does not properly detect the install destination drive for certain (CXI-based) titles. For instance, it is possible for a NAND title to have an SD-looking title ID, e.g. `000400000ff40002` (DevMenu).

Under HOS, this does not matter because the destination is determined by the "is SD app" flag in the ExHeader.

I have reworked the game file installation & title uninstallation process to honor this bit (when applicable). That is:

- TWL titles will always be installed to (Sys/Emu)NAND TWLNAND.
- NCCHs that are CFAs will keep using the original detect-by-titleID method as there is no real way to detect their destination otherwise.
- NCCHs that are CXIs, or containers that have CXIs as their first content (.cxi/.ncch/.app/.cia/cdn contents + tmd) will be installed to the destination indicated by the "is SD app" flag.
- DLC titles and gamecart (image)s will always be installed to (Emu/Sys)NAND SD.
- During title uninstallation, the current mount path and the title ID will be used to determine the uninstallation source path.

I have tested this to be working with all major types of content.

### DIFF
--- a/arm9/source/utils/gameutil.c
+++ b/arm9/source/utils/gameutil.c
@@ -1841,7 +1841,7 @@ u32 UninstallGameDataTie(const char* path, bool remove_tie, bool remove_ticket, 
     u64 tid64;
 
     const char* mntpath = GetMountPath();
-    if (!mntpath) return 1;
+    if (!mntpath || !*mntpath) return 1;
 
     // title.db from emunand?
     if ((strncasecmp(mntpath, "B:/dbs/title.db", 16) == 0) ||
@@ -1861,7 +1861,7 @@ u32 UninstallGameDataTie(const char* path, bool remove_tie, bool remove_ticket, 
     if (is_twl)
         drv[0] = from_emunand ? '5' : '2';
     else
-        drv[0] = GetMountPath()[0];
+        drv[0] = mntpath[0];
 
     return UninstallGameData(drv, tid64, remove_tie, remove_ticket, remove_save);
 }
--- a/arm9/source/utils/gameutil.c
+++ b/arm9/source/utils/gameutil.c
@@ -24,24 +24,24 @@
 u32 GetCbcBlocks(FIL* file, void* buffer, u64 offset, u32 count, u8* titlekey, u8* forced_iv) {
     u8 iv[16] __attribute__((aligned(4)));
     UINT btr;
-    
+
     // sanity, get IV
     if ((count % 0x10) || (offset % 0x10)) return 1; // not possible
     if (forced_iv) memcpy(iv, forced_iv, 0x10);
     else if ((offset < 0x10) || (fvx_lseek(file, offset - 0x10) != FR_OK) ||
         (fvx_read(file, iv, 0x10, &btr) != FR_OK) || (btr != 0x10))
         return 1;
-    
+
     // load data
     if ((fvx_lseek(file, offset) != FR_OK) ||
         (fvx_read(file, buffer, count, &btr) != FR_OK) || (btr != count))
         return 1;
-    
+
     // decrypt
     if (titlekey) {
         setup_aeskey(0x11, titlekey);
         use_aeskey(0x11);
-        cbc_decrypt(buffer, buffer, count / 0x10, AES_CNT_TITLEKEY_DECRYPT_MODE, iv);   
+        cbc_decrypt(buffer, buffer, count / 0x10, AES_CNT_TITLEKEY_DECRYPT_MODE, iv);
     }
 
     return 0;
@@ -231,7 +231,7 @@ u32 LoadTmdFile(TitleMetaData* tmd, const char* path) {
         return 1;
 
     // second part (read full size)
-    if (ValidateTmd(tmd) == 0) {        
+    if (ValidateTmd(tmd) == 0) {
         if (fvx_qread(path, tmd, 0, TMD_SIZE_N(getbe16(tmd->content_count)), NULL) != FR_OK)
             return 1;
     } else if ((ValidateTwlTmd(tmd) == 0) && (getbe16(tmd->content_count) == 1)) {
@@ -310,7 +310,7 @@ u32 LoadTicketForTitleId(Ticket** ticket, const u64 title_id) {
     char* path_bak = NULL;
     strncpy(path_store, GetMountPath(), 256);
     if (*path_store) path_bak = path_store;
-    
+
     // path to ticket.db
     char path_ticketdb[32];
     char drv = *path_store;
@@ -330,7 +330,7 @@ u32 LoadTicketForTitleId(Ticket** ticket, const u64 title_id) {
 
 u32 GetTmdContentPath(char* path_content, const char* path_tmd) {
     // path_content should be 256 bytes in size!
-    
+
     // get path to TMD first content
     static const u8 dlc_tid_high[] = { DLC_TID_HIGH };
 
@@ -904,7 +904,7 @@ u32 VerifyTadFile(const char* path) {
         return 1;
 
     // verify contents
-    u32 content_start = sizeof(TadStub); 
+    u32 content_start = sizeof(TadStub);
     for (u32 i = 0; i < TAD_NUM_CONTENT; i++) {
         u8 hash[32];
         u32 len = align(hdr->content_size[i], 0x10);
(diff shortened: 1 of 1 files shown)


arm9/source/utils/gameutil.c
```diff
@@ -1509,20 +1643,37 @@ u32 CryptGameFile(const char* path, bool inplace, bool encrypt, bool restore) {
     return ret;
 }
 
-u32 GetInstallDataDrive(char* drv, u64 tid64, bool to_emunand) {
+u32 GetTwlInstallDataDrive(char *drv, bool to_emunand) {
```
d0k3, line 1649: This (and below) seems to do the same thing as before. Also, you now have an unreachable return 0 in GetInstallDataDriveGeneric. Does this change really make sense?

arm9/source/utils/gameutil.c
```diff
@@ -1046,6 +1046,140 @@ u32 VerifyTicketFile(const char* path) {
...
+        if ((fvx_qread(path, ncch, 0, sizeof(NcchHeader), NULL) == FR_OK) &&
+            (ValidateNcchHeader(ncch) == 0) && 
+            (!extheader || (fvx_qread(path, extheader, NCCH_EXTHDR_OFFSET, sizeof(NcchExtHeader), NULL) == FR_OK)) &&
+            DecryptNcch(extheader, NCCH_EXTHDR_OFFSET, sizeof(NcchExtHeader), ncch, NULL) == 0) return 0;
+    } else if (filetype & GAME_NCSD) {
+        if ((fvx_qread(path, ncch, NCSD_CNT0_OFFSET, sizeof(NcchHeader), NULL) == FR_OK) &&
+            (ValidateNcchHeader(ncch) == 0) &&
+            (!extheader || ((fvx_qread(path, extheader, NCSD_CINFO_OFFSET + NCCH_EXTHDR_OFFSET, sizeof(NcchExtHeader), NULL) == 0) &&
+            DecryptNcch(extheader, NCCH_EXTHDR_OFFSET, sizeof(NcchExtHeader), ncch, NULL) == 0))) return 0;
+    } else if (filetype & GAME_CDNTMD) {
```
d0k3, line 1140: This makes the code very hard to read (at least for me). As CDN is special anyways (it even needs to show prompts), I suggest moving the CDN related code to a separate function. Does it really make sense to have the CDNTMD related code here, too? If yes, it's okay for that new function to be called from this function.

arm9/source/utils/gameutil.c
```diff
@@ -1046,6 +1046,140 @@ u32 VerifyTicketFile(const char* path) {
...
+            return 1;
+        }
+
+        // decrypt / load NCCH header from first CIA content
+        u32 ret = 1;
+        if (getbe16(cia->tmd.content_count)) {
+            TmdContentChunk* chunk = cia->content_list;
+            if ((getbe64(chunk->size) < sizeof(NcchHeader)) ||
+                (fvx_qread(path, ncch, info.offset_content, sizeof(NcchHeader), NULL) != FR_OK) ||
+                (!extheader || (fvx_qread(path, extheader, info.offset_content + NCCH_EXTHDR_OFFSET, sizeof(NcchExtHeader), NULL) != FR_OK))) {
```
d0k3: If extheader == NULL, we have an autofail for CIA. That can't be right.

arm9/source/utils/gameutil.c
```diff
@@ -1704,8 +1850,17 @@ u32 UninstallGameDataTie(const char* path, bool remove_tie, bool remove_ticket,
...
         return 1;
+    
+    // tid here is in the 00048000... format if it's a TWL title
+    bool is_twl = ((tid64 >> 32) & 0x8000) == 0x8000;
+    char drv[3] = { 0xFF, ':', '\0' };
+    
+    if (is_twl)
+        drv[0] = from_emunand ? '5' : '2';
+    else
+        drv[0] = GetMountPath()[0];
```
d0k3: Unsure if I understand what this is supposed to do. The whole fucntion is meant only for tie, so what cases is that block meant for?

arm9/source/utils/gameutil.c
```diff
@@ -2869,15 +3024,49 @@ u32 BuildCiaFromGameFile(const char* path, bool force_legit) {
...
     u64 filetype = IdentifyFileType(path);
     u32 ret = 0;
 
     // find out the destination
     u64 tid64 = GetGameFileTitleId(path);
-    if (GetInstallDataDrive(drv, tid64, to_emunand) != 0)
-        return 1;
+
+    // CTR filetypes
+    if (filetype & (GAME_CIA | GAME_NCSD | GAME_NCCH | GAME_CDNTMD)) {
```
d0k3, line 3032: This block is most likely the reason why you made 4 functions out of `GetInstallDataDrive`. For this to be more readable (for me) I'd prefer the install location logic to be in a seperate function (like an extended `GetInstallDataDrive`).

d0k3 (changes requested):
Alright, I finally got around to do a proper review. @ZeroSkill1 - thank you for that, it's highly appreciated. Bear with us for a bit, it seems we're a bit slower in reviewing than you are in doing pull requests.

arm9/source/utils/gameutil.c
ZeroSkill1: In order to determine where the CDN content should be installed, it is necessary to read the NCCH header and extheader using this function. I'll separate the CDN code.

arm9/source/utils/gameutil.c
ZeroSkill1: Indeed, should've been `extheader && ....`

arm9/source/utils/gameutil.c
ZeroSkill1: It's true that this is essentially the same as before, at the time I thought perhaps it would be a bit easier with the different drive types separated instead of chaining multiple ternary conditions. I had quite a hard time figuring out which drive is chosen the way it was before.

arm9/source/utils/gameutil.c
ZeroSkill1: That's what I thought too, but it turns out this function is the main one used for uninstalling _anything_, really. As for the drive selection, I've used this "shortcut" of sorts by looking at all possible uninstall locations:
. | SysNAND CTRNAND | SysNAND TWLNAND | SysNAND SD | EmuNAND CTRNAND | EmuNAND TWLNAND | EmuNAND SD
-- | -- | -- | -- | -- | -- | --
Title path | T:/... | T:/00048... | T:/... | T:/... | T:/00048... | T:/...
Mount path | 1:/dbs/title.db | 1:/dbs/title.db | A:/dbs/title.db | 4:/dbs/title.db | 4:/dbs/title.db | B:/dbs/title.db
Uninstall drive | 1:/ | 2:/ | A:/ | 4:/ | 5:/ | B:/

For all uninstall locations other than for TWL titles, the mountpath of the mounted .db file is the same as the uninstall drive, and for TWL titles, the uninstall drive is the equivalent TWLNAND drive depending on SysNAND/EmuNAND.

I can add a comment explaining this briefly.

d0k3: Thanks, that explains it, I've got one last round of nitpicking, though. I'm currently wondering if it is ever possible for  GetMountPath()[0] to be zero here, and to be blunt, I don't know. NULL for GetMountPath() should be impossible here, but maybe it makes sense to guard an empty string here (because that's a plausible return value here). Other than that, this one is good to be merged 👍

ZeroSkill1: That NULL check is already performed earlier in the function. I guess I could just use the existing `mntpath` instead of using `GetMountPath()` again, and use a simple empty string check in the earlier check.

d0k3:
All good, thank you @ZeroSkill1 !

PR #957, opened 2026-03-30
Italian: Fix newlines in decimal and thousand separators

Epicpkmn11:
Fixed on crowdin side too, this is the only glaring problem I see with the new build


Epicpkmn11:
Obsoleted by e53b03b854cfb9a6fe8e90099c1077810edbcc3a

Issue #726, opened 2021-08-26
[FEATURE REQUEST] Ability to Decrypt/Encrypt 3DS Game Cartridge .sav Files


DeadSkullzJr:
**Got a great idea on how to improve GodMode9?**
Add the ability to decrypt and encrypt raw .sav files from 3DS game cartridges.

**Describe the feature you'd like**
GodMode9 is able to dump encrypted .sav files, but you can't decrypt the data with the current builds. I would like the ability to decrypt and encrypt the .sav files, allowing the ability to directly edit the decrypted .sav file instead of splitting it into pieces using homebrew alternatives such as JK's Save Manager or Checkpoint. Afterwards, you could then encrypt it and import that save back into the 3DS game cartridge using whatever preferred solution (unless GodMode9 gets the ability to import .sav files all on its own of course).

**Describe alternatives you've considered**
I've used both JK's Save Manager and Checkpoint, while their purpose is solid, I prefer the ability to maintain my own .sav files as a whole and not just portions of it.

**What I am NOT asking for**
I am not asking for something like JKSM or Checkpoint, those pieces of software don't produce any type of .sav files, instead those two produce save data in pieces (not that it's a bad thing, it's just not what I am looking for at the moment). They also tend to alter information after importing data even if the data imported was unaltered in any way, which to me doesn't make sense given if the data is unaltered, it should yield 1:1 results from a save backup, that just makes me think there are some inaccuracies or inconsistencies somewhere. I personally value my save data, I feel we need a better solution.

**Side Note**
I tried looking into a P.C. software based solution to handle save encryption and decryption given that's faster than using the Nintendo 3DS, but stumbled on nothing viable of the sorts. I don't know 100% how save decryption works exactly with 3DS game cartridges, all I know is you can't really mix .sav files UNLESS it was already decrypted and then re-encrypted specifically for your game cartridge, which to me tells me it has something to do with unique identifiers of the game cartridge while still using the same encryption algorithm, might be something that could be brute forced one day with special software. There isn't a whole heck of a lot of research on preserving .sav integrity for the 3DS, just mostly ways to split it into pieces, then importing amalgamated Frankenstein data, which is fine and dandy, but I want that type of control at the raw .sav level.

[...]

PR #963, opened 2026-05-02
[Testing needed] Add support for decrypting gamecart saves

### LINKED ISSUE #958: [FEATURE REQUEST] Add Cart ID 2 to cart dump metadata .txt
Cart ID 2 should be added to the cart dump metadata .txt file, like Cart ID. This, like with the Cart ID functionality, makes it easy for this information to be preserved when people dump carts (and submit dump info to preservation databases).

(Functionality to retrieve Cart ID 2 was recently added to GodMode9 https://github.com/d0k3/GodMode9/pull/862)

N/A

ZeroSkill1:
This adds support for decrypting and reading the inner DISA save image stored on gamecarts.

For gamecarts using SPI flash for save data, this also processes the underlying wear leveling layer (that is, the blockmap and journal) and reconstructs the save image accordingly.

All possible crypto methods should be supported (but not all have been tested yet, see below).

Currently, this feature only provides read access to the decrypted save data. (would be half of what #726 is asking for)

The decrypted save image is provided as an additional file in the gamecart (`C:`) drive, as `{title_id}_v{version}.dec.sav`. For carts using wear leveling, the decrypted save image will be smaller than the raw version by exactly two flash sectors (4*2 KiB) due to the wear leveling blockmap and journal (and its backup copy) being stored in it. For CARD2 carts both versions will be equal in size because those carts do not implement wear leveling.

In addition, these changes have been made to the gamecart code:
- Any mounted image (.nds, .3ds, decrypted save for 3DS cartridges) will be unmounted if the cart is physically ejected (as opposed to the "dead" remaining drive and files inside it after the cart has been removed)
- Card ID2 is now also provided in the `.txt` file of the gamecart drive for 3DS cartridges (fixes #958)
- Removed an unnecessary spam of command A2h during secure init - this improves init times by up to **2.5s**

Example (Super Mario 3D Land):

Example showing decrypted savegame in gamecart menu.

More testing is required as I do not personally have all possible types of cartridges. Here are the ones I have tested:

- Zoo Resort 3D (000400000004B200) - CARD1 - SPI Flash (128 KiB) - Crypto Type 0
- Super Mario 3D Land (0004000000053F00) - CARD1 - SPI Flash (128 KiB) - Crypto Type 1

Feel free to share test results. Working means you can mount the DISA image and see plaintext save data (or visible `DISA`, `DIFI`, `DPFS`, and `IVFC` magic in the hex editor). Please include the `.txt` file GodMode9 generates in the gamecart drive when doing so. If something doesn't work right, please also provide the `ncsd.bin` of your cartridge (can be extracted from the .3ds file by mounting it - only the last 0x100 bytes of the whole 0x200 bytes are necessary in this case).

The following cart types need testing (checked entries mean these have been tested to be working by other users):

- [x] 512 KiB SPI Flash
- [x] 1 MiB SPI Flash
- [x] Carts using crypto type 2 (used in games such as Pokémon Omega Ruby and Alpha Sapphire)
- [x] New3DS-only carts using crypto type 2 (used in the physical version of Runbow Pocket and possibly some other New3DS exclusives)
- [x] CARD2 (also used in above games)


d0k3:
@ZeroSkill1 - I just noticed you already did #958 with that pull request. I won't commit mine then, I guess. For reviewing that one, I'll wait until testing is finished.

arm9/source/game/ncsd.h
```diff
@@ -33,7 +33,8 @@ typedef struct {
     u8  sector_zero_offset[0x4];
     u8  partition_flags[8];
     u8  partitionId_table[8][8];
-    u8  reserved[0x30];
+    u8  reserved[0x2F];
+    u8  extra_save_keysel;
```
d0k3, line 37: Is that byte documented somewhere? Just asking to make sure.

arm9/source/crypto/aes.h
```diff
@@ -40,6 +40,7 @@ extern "C" {
 #define AES_CNT_FLUSH_WRITE   0x00000400u
 
 #define AES_CNT_CTRNAND_MODE (AES_CTR_MODE | AES_CNT_INPUT_ORDER | AES_CNT_OUTPUT_ORDER | AES_CNT_INPUT_ENDIAN | AES_CNT_OUTPUT_ENDIAN)
+#define AES_CNT_CART_SAVE_MODE (AES_CTR_MODE | AES_CNT_INPUT_ORDER | AES_CNT_OUTPUT_ORDER | AES_CNT_INPUT_ENDIAN | AES_CNT_OUTPUT_ENDIAN)
```
d0k3: This is exactly the same thing as AES_CNT_CTRNAND_MODE. Does it make sense to have a separate macro for that? Not against it, but maybe a comment would help.

arm9/source/gamecart/protocol.c
```diff
@@ -211,10 +211,10 @@ void Cart_Secure_Init(u32 *buf, u32 *out)
         CTR_SendCommand(C5_cmd, 0, 1, 0x100002C, NULL);
     }
 
-    for (int i = 0; i < 5; ++i) {
-        CTR_SendCommand(A2_cmd, 4, 1, 0x701002C, &test);
-        ARM_WaitCycles(0xF0000 * 8);
-    }
+    //for (int i = 0; i < 5; ++i) {
```
d0k3, line 216: To my understanding, this one was not needed. Still, it would help if the removal was explained in a comment.

arm9/source/godmode.c
```diff
@@ -3079,8 +3080,13 @@ u32 GodMode(int entrypoint) {
             if (!InitVCartDrive() && (pad_state & CART_INSERT) &&
                 (curr_drvtype & DRV_CART)) // reinit virtual cart drive
                 ShowPrompt(false, "%s", STR_CART_INIT_FAILED);
-            if (!(*current_path) || (curr_drvtype & DRV_CART))
+            // unmount any gamecart image if the gamecart was removed
```
d0k3, line 3082: Just out of pure interest: what really happens when you have something mounted from the gamecart, and then eject it? Does the fix work? In any case, doing this here makes sense.

arm9/source/godmode.c
```diff
@@ -1,6 +1,7 @@
 #include "godmode.h"
 #include "paint9.h"
 #include "memmap.h"
+#include "protocol.h"
```
d0k3: What uses the additional include here?

arm9/source/gamecart/gamecart.h
```diff
@@ -41,4 +174,5 @@ u32 ReadCartBytes(void* buffer, u64 offset, u64 count, CartData* cdata, bool car
 u32 ReadCartPrivateHeader(void* buffer, u64 offset, u64 count, CartData* cdata);
 u32 ReadCartInfo(u8* buffer, u64 offset, u64 count, CartData* cdata);
 u32 ReadCartSave(u8* buffer, u64 offset, u64 count, CartData* cdata);
+u32 ReadDecryptedCartSave(u8* buffer, u64 offset, u64 count, CartData* cdata);
```
d0k3: This function is the only really required addition here. You'd possibly need another function for  the wear leveling thing, but you don't need to expose all these new (and old) structs in a header file that's not meant to include these kind of details.

arm9/source/gamecart/gamecart.c
```diff
@@ -160,6 +132,245 @@ static u32 GetCtrCartSaveSize(CartData* cdata) {
     }
 }
 
+void SetBlockmapEntry(u32 index, SaveBlockmapEntry *in_entry, CartWearLevelingData *wldata) {
```
d0k3: About this whole file - these changes are too much for a file that's basically just meant to provide a simple interface to  the complex functionality found in this source file folder. For better readability, I suggest moving all the complex stuff to a separate source file and to leave the changes in here to a minimum. One or two simple interface functions may be enough.

d0k3 (changes requested):
I figured I can already do the review while we wait for testing. Again, thanks for your contribution!

Additional, to these comments, I wonder what happens if the save does not include a valid DISA magic. Will there be problems?

arm9/source/game/ncsd.h
ZeroSkill1: Yes: https://3dbrew.org/wiki/NCSD#NCSD_header. [This section](https://3dbrew.org/wiki/NCSD#Partition_Flags_(In_Terms_of_Save_Crypto_Determination)) provides some more context as for what the byte does.

arm9/source/crypto/aes.h
ZeroSkill1: I was having trouble with data not decrypting properly unless I set the endianness options, and using `AES_CNT_CTRNAND_MODE` in gamecart code didn't make much sense to me. ~~I'll add a comment.~~ I suppose I could also just define `AES_CNT_CART_SAVE_MODE` to be `AES_CNT_CTRNAND_MODE`.

arm9/source/godmode.c
ZeroSkill1: Before, when the gamecart was removed, the mounted game image (NCSD) or mounted save data (DISA) would remain in the drive list despite the mounted source file in the gamecart drive no longer being valid. With this fix, the "zombie" `NCSD GAME IMAGE` or `DISA/DIFF IMAGE` drives are removed automatically.

arm9/source/godmode.c
ZeroSkill1: Actually, nothing; this seems to be a remnant from when I was testing card init related things.

arm9/source/godmode.c
al3x10m: i think without it it would say smth like "invalid path" or smth...but this is nice too :3

ZeroSkill1:
I've restructured the relevant code as you suggested and now both save crypto and wear leveling are in a separate file. Looking at the code now, it doesn't make sense for any of my changes to be in gamecart.c, as that file clearly is meant for raw, device-level reads/writes, while this new feature is entirely composed of higher-level stuff.

> I wonder what happens if the save does not include a valid DISA magic. Will there be problems?

This *can* happen, but only if the save data is unformatted. There would be two cases:

1. For CARD1, if the wear leveling data exists, or the cart is CARD2 (which doesn't use wear leveling) and the encrypted inner contents are blank (no DISA). In this case, the .dec.sav becomes available to read from raw, and of course the DISA/DIFF code won't be able to mount it.
2. For CARD1 only: If the wear leveling data does not exist at all (e.g. if the save chip is completely FFs). In this case, the .dec.sav becomes unavailable completely because all CARD1 carts must have wear leveling data for the inner contents to be usable.

TL;DR: no DISA magic means no DISA mounting, and for CARD1, no wear leveling data means no `.dec.sav`.

arm9/source/gamecart/save_ctr.c
```diff
@@ -0,0 +1,489 @@
...
+    return 0;
+}
+
+int InitCtrCardSave(CartData *cdata) {
+    memset(&savectx, 0, sizeof(savectx));
+
+    // the wear leveling header exists twice:
+    // the one at 0x0 is the main one
+    // the one at 0x1000 is used as a failsafe if the one above is corrupt
+    if (InitSaveWearLeveling(cdata, 0) != 0 && InitSaveWearLeveling(cdata, 0x1000) != 0)
```
d0k3: If the first `InitSaveWearLeveling` call fails, savectx may be partially written before the fallback call starts. A memset between the two would be safer.

arm9/source/gamecart/save_ctr.c
```diff
@@ -0,0 +1,489 @@
...
+
+    u32 cmac_keyslot = savectx.crypto_type == CARD_SAVE_CRYPTO_V1_N3DS ? CARD_SAVE_CMAC_KEYSLOT_N3DS : CARD_SAVE_CMAC_KEYSLOT_O3DS;
+    u32 crypto_keyslot = savectx.crypto_type == CARD_SAVE_CRYPTO_V1_N3DS ? CARD_SAVE_CRYPTO_KEYSLOT_N3DS : CARD_SAVE_CRYPTO_KEYSLOT_O3DS;
+
+    setup_aeskeyY(cmac_keyslot, save_key_y); // savedata CMAC key
+    setup_aeskeyY(crypto_keyslot, save_key_y); // savedata crypto key
+    return 0;
+}
+
+int InitCtrCardSave(CartData *cdata) {
```
d0k3: That's just a minor style thing, but shouldn't this be u32?

d0k3 (changes requested):
Thanks for the last changes, lookng much better now. So, last two comments - as everything is tested now, after that this one is good to go as well.

arm9/source/gamecart/save_ctr.c
ZeroSkill1: Indeed. I only later saw most returns are u32 across the codebase.

d0k3 (approved):
All good, this one is ready to merge.

PR #964, opened 2026-05-10
Implement savedata/extdata mounting & missing .db files creation

### LINKED ISSUE #731: [FEATURE REQUEST] Add support for mounting SAVE and VSXE inner fat filesystems
**Got a great idea on how to improve GodMode9?**
I'm mainly making this issue to consolidate discussion about this and to pressure myself to actually do it. This is something I've wanted to work on adding since I added BDRI support and just haven't. Though, if someone else reads this and decides to implement it before I get to doing so, I'm not complaining.

Support for GodMode9 to mount .sav DISA files (SAVE filesystem) and groups of DIFF extdata files (VSXE filesystem) as a virtual drive, similarly to how it can currently mount .db DIFF files (BDRI filesystem). Again, I'm planning to implement this myself, and am just making this issue to organize things.

Can't think of any better/other way to implement this besides expanding the current functionality to mount the other inner FAT variants. One thing to note is that gm9 *kinda* already supports viewing and modifying extdata files, just by mounting the individual DIFF containers that correspond to each file in the VSXE filesystem, but this can require some trial and error for a user to figure out which file is which and is almost useless from a scripting perspective. This also does not allow adding or deleting files, which is something I would like to support.

ZeroSkill1:
# Save data / Extdata mounting
Fixes #731.

Implements mounting save data (SAVE) and extdata (EXSV) inside DISA and DIFF images respectively, allowing users to explore the actual contents of these images rather than just the raw partitionA/B.bin.

- For save data: Whenever a DISA image is mounted and contains a valid SAVE filesystem, the `[F:] SAVE FILE IMAGE` drive is added, which leads to the inner contents of the image.
- For extdata: Whenever the DIFF file `00000000/00000001` inside an extdata directory is mounted and contains a valid EXSV file system, the `[F:] EXTDATA IMAGE` drive is added, functioning similar to the above mentioned.

This is especially useful for extracting savedata/extdata for games, or extracting/exploring system savedata and/or shared NAND extdata.

Currently only read access is supported, but in the future I plan to add write support as well.

Example mounting the CEC (StreetPass) system save data and exploring its contents:
Example mounting the CEC (StreetPass) system save data and exploring its contents.

Example having mounted the shared NAND extdata `0xf000000b`:
grafik

# Missing .db files creation

Adds support for creating the various .db files (except certs.db) completely from scratch, enabling title installation even without booting into HOS before that. The user is able to check for and recreate missing .db files by opening the "context menu" of the respective drive (R+A) and choosing the option `Create missing .db files`. If any of the files are missing, the user will be prompted to unlock write permissions, and all possible .db files for that drive will be created. If no files are missing, the user is informed accordingly:

grafik

This new functionality supports recreating the following .db files according to the target drive:
- SysNAND/EmuNAND SD: `import.db`, `title.db`
- SysNAND/EmuNAND CTRNAND: `import.db`, `title.db`, `ticket.db`, `tmp_i.db`, `tmp_t.db`

`certs.db` is not covered by this new feature, as it uses a different format. Recreating it from scratch would be pointless without the default certificates that are expected in both GM9 and HOS.

I have also added this new feature to the Lua API, exposed as `fs.create_dbs`, allowing for fully HOS-less automated title installation scripts.


arm9/source/game/saveexsv.c
```diff
@@ -0,0 +1,258 @@
...
+                                    sav->fs_info.dir_hashtbl.count * sizeof(u32) +
+                                    sav->fs_info.file_hashtbl.count * sizeof(u32) +
+                                    (sav->fs_info.fat.count + 1) * sizeof(SaveFatEntry) +
+                                    dir_table_size +
+                                    file_table_size,
+                                    sav->pre_header.fs_image_blocksize);
+        u64 min_part_b_size = sav->fs_info.data_region.count * sav->fs_info.data_region_blocksize;
+                              
+        if (part_a_size < min_part_a_size || part_b_size < min_part_b_size)
+            return 1;
```
d0k3: No err_exit in that case. Is that intended or is it a potential resource leak? If intended an explaining comment would help.

arm9/source/game/saveexsv.c
```diff
@@ -0,0 +1,258 @@
...
+
+    if (read_alloc_chunk((void **)&sav->dir_hashtbl, &part_a, sav->fs_info.dir_hashtbl.offset, sav->fs_info.dir_hashtbl.count * 4) ||
+        read_alloc_chunk((void **)&sav->file_hashtbl, &part_a, sav->fs_info.file_hashtbl.offset, sav->fs_info.file_hashtbl.count * 4) ||
+        read_alloc_chunk((void **)&sav->fat_entries, &part_a, sav->fs_info.fat.offset, (sav->fs_info.fat.count + 1) * sizeof(SaveFatEntry)) ||
+        read_alloc_chunk((void **)&sav->dir_entries, &part_a, dir_table_offset, dir_table_size) ||
+        read_alloc_chunk((void **)&sav->file_entries, &part_a, file_table_offset, file_table_size)) {
+        goto err_exit;
+    }
+
+    sav->init_ok = true;
```
d0k3: That's dead code. The variable get's immediately overwritten below.

arm9/source/game/saveexsv.c
```diff
@@ -0,0 +1,258 @@
...
+ret:
```
d0k3, line 137: To be honest, I never liked goto logic much. It's okay in that case (quite easily readable, too), but don't make that a habit :)

arm9/source/game/saveexsv.c
```diff
@@ -0,0 +1,258 @@
...
+
+    return 0;
+}
+
+int SaveExsvReadFatFile(SaveExsvFile *sav, u32 index, void *buffer, u32 offset, u32 count) {
+    if (!sav->init_ok || index >= sav->max_num_file_entries)
+        return 1;
+
+    if (sav->is_exsv) {
+        return 0;
```
d0k3: This is more likely me not really understanding what's going on rather than a real issue. If the silent success here is okay, maybe you could add a short comment explaining it.

arm9/source/virtual/vsaveexsv.c
```diff
@@ -0,0 +1,183 @@
...
+}
+
+u64 InitVSaveDrive(void) { // prerequisite: save mounted as virtual disa image
+    u64 mount_state = CheckVDisaDiffDrive();
+    if (!(mount_state & (SYS_DISA | SYS_DIFF))) return 0;
+
+    DeinitVSaveDrive();
+
+    if (SaveExsvFileInit(&sav))
+        DeinitVSaveDrive();
```
d0k3, line 41: That may be another case of me not getting it, but is it save to continue after the error case? ´sav´ may have been written to in that case, with unsure consequences. Maybe there's a `return` missing here?

arm9/source/virtual/vsaveexsv.c
```diff
@@ -0,0 +1,183 @@
...
+}
+
+u64 CheckVSaveDrive(void) {
+    u64 mount_state = CheckVDisaDiffDrive();
+    return ((mount_state & (SYS_DISA | SYS_DIFF)) && sav.init_ok) ? mount_state : 0;
+}
+
+void DeinitVSaveDrive() {
+    SaveExsvFileFree(&sav);
+    fvx_close(&part_a);
```
d0k3: We don't check if the handle was open in that case and below. May be better to guard that. Besides, are these handles ever opened in this file?

arm9/source/virtual/vsaveexsv.c
```diff
@@ -0,0 +1,183 @@
...
+            vdir->index = 2;
+        }
+
+        return true;
+    }
+
+    if (vdir->index == 2) {
+        SaveDirectoryEntry *dir = &sav.dir_entries[vdir->offset];
+        if (!dir->ent.first_subdir_index)
+            return false;
```
d0k3, line 123: Only a suspicion, but is it possible for this to skip pending sibling directories?

arm9/source/game/saveexsv.c
```diff
@@ -0,0 +1,258 @@
...
+        return 1;
+    }
+    return 0;
+}
+
+static FIL part_a = { 0 }, part_b = { 0 };
+static FIL *data_part = NULL;
+
+int SaveExsvFileInit(SaveExsvFile *sav) {
+    UINT nread = 0;
```
d0k3, line 27: Unsure about that, but shouldn't you check if `sav->init_ok` is already set before doing everything else in that function?

arm9/source/virtual/vsaveexsv.c
```diff
@@ -0,0 +1,183 @@
...
+        return 0;
+
+    if (sav.is_exsv) {
+        return 0; // maybe implement Quota.dat reading for this at some point
+    } else {
+        return sav.pre_header.fs_image_size_blocks * sav.pre_header.fs_image_blocksize;
+    }
+}
+
+bool VSaveIsExtData() {
```
d0k3: Nitpick, but to be consistent, the parameter should be ´(void)´

arm9/source/virtual/vsaveexsv.h
```diff
@@ -0,0 +1,13 @@
...
+
+#include "virtual.h"
+
+void DeinitVSaveDrive(void);
+u64 InitVSaveDrive(void);
+u64 CheckVSaveDrive(void);
+
+bool ReadVSaveDir(VirtualFile* vfile, VirtualDir* vdir);
+int ReadVSaveFile(const VirtualFile* vfile, void* buffer, u64 offset, u64 count);
+bool VSaveIsExtData();
```
d0k3: See above, parameter should be ´(void)´

arm9/source/game/saveexsv.h
```diff
@@ -0,0 +1,167 @@
...
+    SaveFsInfo fs_info;
+    SavePreHeader pre_header;
+    u32 *dir_hashtbl;
+    u32 *file_hashtbl;
+    SaveFatEntry *fat_entries;
+    SaveDirectoryEntry *dir_entries;
+    u32 max_num_dir_entries;
+    SaveFileEntry *file_entries;
+    u32 max_num_file_entries;
+    u8 *block_buffer;
```
d0k3: Is that ever used, or is it a leftover? :)

d0k3 (changes requested):
Review done. If I haven't mentioned that yet, thanks for the new feature!

arm9/source/game/saveexsv.c
ZeroSkill1: Indeed, should be an err_exit case.

arm9/source/game/saveexsv.c
ZeroSkill1: Me neither - but in this case, there is so much that can go wrong and adding the same cleanup routine for every little thing can get overwhelming quick.

arm9/source/game/saveexsv.c
ZeroSkill1: It should actually be a silent failure. Must have gotten that confused with another part of the codebase, because I remember some parts have failure and success reversed. Just to clarify, extdata doesn't use the file FAT unlike SAVE - so reading an extdata file this way should not ever happen, because extdata always uses numbered physical files. I'll add a short comment explaining this.

arm9/source/virtual/vsaveexsv.c
ZeroSkill1: Yep, missing return.

arm9/source/virtual/vsaveexsv.c
ZeroSkill1: Those are remnants from when the files were managed in the virtual drive code. I'll remove them.

arm9/source/virtual/vsaveexsv.c
ZeroSkill1: Sibling directories shouldn't be a concern here, because this always lists the parent directory, starting from root. Therefore, only subfiles and subdirs of that parent directory are processed.

arm9/source/game/saveexsv.c
ZeroSkill1: Yes, but only if it's possible for this (and therefore InitVSaveDrive) to be called more than once consecutively. Not sure if that can happen, but I'll test and see.

arm9/source/game/saveexsv.h
ZeroSkill1: It (`block_buffer`) is a leftover from an older version of the SAVE code I had lying around.

arm9/source/game/saveexsv.c
ZeroSkill1: Tested it, and it doesn't seem like that can ever happen. I'll add a short-circuit return though just in case it somehow does.

ZeroSkill1:
Sorry to add even more to this already massive PR, but .db files creation will also be included in this one. I shouldn't have touched anything that affected most of the code already reviewed (I hope). The inner-FAT code needed a little restructuring anyway that I felt would be better off also being included here.

arm9/source/game/saveexsv.c
```diff
@@ -0,0 +1,261 @@
...
+    fvx_close(&part_b);
+    sav->init_ok = !retval;
+    return retval;
+err_exit:
+    retval = 1;
+    goto ret;
+}
+
+void SaveExsvFileFree(SaveExsvFile *sav) {
+    fvx_close(&part_a);
```
d0k3: Is that required? I think both handles are already closed at that point.

d0k3:
> Sorry to add even more to this already massive PR, but .db files creation will also be included in this one. I shouldn't have touched anything that affected most of the code already reviewed (I hope). The inner-FAT code needed a little restructuring anyway that I felt would be better off also being included here.

It's alright, reviewing that will just take me a bit.

arm9/source/game/disadiff.c
```diff
@@ -570,4 +620,197 @@ u32 WriteDisaDiffIvfcLvl4(const char* path, const DisaDiffRWInfo* info, u32 offs
...
+    if (fvx_open(&fp, path, FA_WRITE | FA_OPEN_EXISTING) != FR_OK ||
+        fvx_size(&fp) < CalcDiffFileSize(&diff) ||
+        (fvx_write(&fp, cmac, sizeof(cmac), &written) != FR_OK || written != sizeof(cmac)) ||
+        (fvx_write(&fp, &diff, sizeof(diff), &written) != FR_OK || written != sizeof(diff)) ||
+        fvx_lseek(&fp, diff.offset_table1) != FR_OK ||
+        (fvx_write(&fp, part_desc, diff.size_table, &written) != FR_OK || written != diff.size_table) ||
+        fvx_lseek(&fp, diff.offset_table0) != FR_OK ||
+        (fvx_write(&fp, part_desc, diff.size_table, &written) != FR_OK || written != diff.size_table)) {
+        free(part_desc);
+        fvx_close(&fp);
```
d0k3, line 813: If fvx_open fails, fvx_close is done on an undefined handle.

arm9/source/game/bdri.c
```diff
@@ -898,3 +917,222 @@ u32 AddTicketToDB(const char* path, const u8* title_id, const Ticket* ticket, bo
...
+            break; // found a satisfactory number of data blocks
+
+        --num_data_blocks;
+    }
+
+    // create file system info
+
+    fs_info->data_region_blocksize = blocksize;
+
+	fs_info->dir_hashtbl.outfat_offset = pre_header->fs_info_offset + sizeof(IFatFsInfo);
```
d0k3: This and below - you're using tab indentations instead of space (which is used everywhere else in the file).

arm9/source/game/bdri.c
```diff
@@ -898,3 +917,222 @@ u32 AddTicketToDB(const char* path, const u8* title_id, const Ticket* ticket, bo
...
+	fs_info->dirtable_info.starting_block_index = 0xFFFFFFFF; // to be allocated in FAT later using above block_count
+
+	fs_info->filetable_info.max_entry_count = num_files;
+	fs_info->filetable_info.block_count = file_entries_num_blocks;
+	fs_info->filetable_info.starting_block_index = 0xFFFFFFFF; // to be allocated in FAT later using above block_count
+
+	FIL file;
+
+    if (fvx_open(&file, path, FA_READ | FA_WRITE | FA_OPEN_EXISTING) != FR_OK)
+        return 1;
```
d0k3, line 1019: It may make sense to also set bdrifp to NULL in that error case.

arm9/source/utils/gameutil.c
```diff
@@ -4360,3 +4360,175 @@ u32 GetGoodName(char* name, const char* path, bool quick) {
 
     return 0;
 }
+
+typedef struct DbFileConfig {
```
d0k3: `gameutil.c` is already pretty big, we need to pay attention to use the file only for stuff that fits. I think the new additions better fit into `bdri.c`.

arm9/source/filesys/fsdrive.c
```diff
@@ -123,6 +124,8 @@ bool GetRootDirContentsWorker(DirStruct* contents) {
             snprintf(entry->name, 252, "[%s] %s (%s)", drvnum[i], drvname[i], carttype);
         else if (*(drvnum[i]) == '0') // SD card handling
             snprintf(entry->name, 252, "[%s] %s (%s)", drvnum[i], drvname[i], sdlabel);
+        else if (*(drvnum[i]) == 'F') // Save/Extdata handling
+            snprintf(entry->name, 252, "[%s] %s", drvnum[i], VSaveIsExtData() ? STR_LAB_EXTDATA_IMAGE : STR_LAB_SAVE_FILE_IMAGE);
```
d0k3: This is better done similar to the `G:` drive handling above. We don't need the `vsaveexsv.h` include then.

arm9/source/game/saveexsv.c
ZeroSkill1: Yep, looks like that's unecessary. The files are closed at the end of SaveExsvFileInit.

ManiacOfGitHub:
How easy would reading the config savegame time offset be after this to make nullification unnecessary?

ZeroSkill1:
Should just be a matter of mounting the config save and reading the offset. Definitely possible.

arm9/source/game/bdri.c
```diff
@@ -348,148 +356,163 @@ static u32 RemoveBDRIEntry(const BDRIFsHeader* fs_header, const u32 fs_header_of
...
-                return 1; // Reached the end of the free node chain without finding enough contiguous free space - this should never realistically happen
+        if (node_size - size_blocks > 1) {
+            if (BDRIRead(fat_offset + (fat_index + node_size - 1) * FAT_ENTRY_SIZE, FAT_ENTRY_SIZE, fat_entry) != FR_OK)
+                return 1;
 
-            fat_index = next_fat_index;
+            fat_entry[0] = buildfatuv(fat_index + size_blocks, true);
 
-            if (BDRIRead(fat_offset + fat_index * FAT_ENTRY_SIZE, FAT_ENTRY_SIZE, fat_entry) != FR_OK)
+            if ((BDRIWrite(fat_offset + (fat_index + node_size - 1) * FAT_ENTRY_SIZE, FAT_ENTRY_SIZE, fat_entry) != FR_OK) ||
```
d0k3, line 415: Here, the new second entry of the leftover free node is written using the old tail's data. But the tail's u[1] is just the terminator {0, false} — it doesn't point anywhere. So the new second entry loses the actual tail index, and the node's size can't be computed correctly afterwards. Only shows up when leftover > 2 blocks, which is basically guaranteed on the first allocation into a fresh DB. Pretty sure this breaks the free list for any of the bigger DBs (title.db etc.).

arm9/source/game/bdri.c
```diff
@@ -898,3 +923,396 @@ u32 AddTicketToDB(const char* path, const u8* title_id, const Ticket* ticket, bo
...
+        return 1;
+    }
+
+    // initialize file/dir hash tables & entries
+
+    // the directory hashtable has just one entry, pointing to the root directory
+    // no need to do the calculation
+    u32 dir_hash_value = 1;
+
+    if (BDRIWrite(image_offset + fs_info->dir_hashtbl.outfat_offset, sizeof(u32), &dir_hash_value) != 0) {
```
d0k3: BDRIWrite(...) != FR_OK would be more consistent here.

arm9/source/game/bdri.c
```diff
@@ -898,3 +923,396 @@ u32 AddTicketToDB(const char* path, const u8* title_id, const Ticket* ticket, bo
...
+    memset(zero, 0, sizeof(zero));
+
+    u32 remaining_fht = file_hashbucket_count * 4;
+    u32 fht_cur_offset = fs_info->file_hashtbl.outfat_offset;
+
+    // fill the file hashtable with zeros
+    while (remaining_fht) {
+        u32 writesize = min(remaining_fht, sizeof(zero));
+
+        if (BDRIWrite(image_offset + fht_cur_offset, writesize, zero) != 0) {
```
d0k3: See above - BDRIWrite(...) != FR_OK would be more consistent here.

arm9/source/game/bdri.c
```diff
@@ -898,3 +923,396 @@ u32 AddTicketToDB(const char* path, const u8* title_id, const Ticket* ticket, bo
...
+    SDMC_TITLE_DB, SDMC_IMPORT_DB,
+} DbType;
+
+static const DbFileConfig DbConfigs[7] = {
+    // NAND dbs
+    { "/dbs/ticket.db", "TICK"   , 1, 0x10A2010, 0x10, 0x10A2000, 0x200, 8192 }, // ticket.db is the only db where the image size is correct...
+    { "/dbs/import.db", "NANDIDB", 0, 0x17800  , 0x80, 0x17800  , 0x80 , 512  }, // ---
+    { "/dbs/title.db" , "NANDTDB", 0, 0x2ED80  , 0x80, 0x2ED80  , 0x80,  1024 }, //   | *all* of these are missing 0x80 bytes at the end.
+    { "/dbs/tmp_i.db" , "TEMPIDB", 0, 0xE00    , 0x80, 0xE00    , 0x80,  16   }, //   | the image size should be file size - 0x80 because
+    { "/dbs/tmp_t.db" , "TEMPIDB", 0, 0x6000   , 0x80, 0x6000   , 0x80,  128  }, //   | of the pre-header. keeping it the same just for
```
d0k3, line 1170: Is "TEMPIDB" correct here? I'm asking because I don't know :^).

d0k3 (changes requested):
I'm, sorry, I found some more stuff. It's a big pull request after all. I'm fairly sure I won't be able to find any more problems after this, though.

arm9/source/game/bdri.c
ZeroSkill1: Despite the weird naming, yes: https://www.3dbrew.org/wiki/Title_Database#Database_Magic

arm9/source/game/bdri.c
ZeroSkill1: I didn't actually change this part of the code, other than extracting it out of `AddBDRIEntry` to allocate the file/dir entry tables with.

Anyway, according to [the existing documentation](https://3dbrew.org/wiki/Inner_FAT#File_Allocation_Table), a node's tail entry points to **itself** (via `index_V`, assuming that is what you meant by `u[1]` of tail), not a `{0, false}` terminator. This is also observed in the [freelist formatting code in CreateBDRI](https://github.com/ZeroSkill1/GodMode9/blob/save-mounting/arm9/source/game/bdri.c#L1045-L1049).

Therefore, the tail index _should_ be preserved, since it is present twice in any given node (including the freelist): once in `node[1]` and again in the tail, `node[n - 1]` for a node spanning `n` entries.

d0k3:
Merged, thanks a lot (not only for this work, but also for your patience!)

PR #967, opened 2026-06-18
Lua: Add UTF16 <-> UTF8 conversion

Gruetzig:
Helpful when interacting with system files that save in UTF16 format.

### DIFF
--- a/arm9/source/lua/gm9internalutil.c
+++ b/arm9/source/lua/gm9internalutil.c
@@ -1,0 +1,38 @@
+#ifndef NO_LUA
+#include "gm9internalutil.h"
+#include "utf.h"
+#include "fs.h"
+#include "ui.h"
+
+static int internalutil_utf16_to_utf8(lua_State* L) {
+    CheckLuaArgCount(L, 1, "_util.utf16_to_utf8");
+    size_t strInLength = 0;
+    const u16* strIn = (const u16*)(void*)luaL_checklstring(L, 1, &strInLength);
+    u8 buffer[2*strInLength]; //assume a maximum of 4 utf8 codepoints per utf16 codepoint (2 bytes per utf16 codepoint so * 2)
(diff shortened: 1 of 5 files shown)


d0k3:
@ihaveamac - I think I will need your input on that one.

data/luapackages/util.lua
```diff
@@ -47,4 +47,13 @@ function util.running_as_module()
     return type(required) == "string"
 end
 
+function util.utf16_to_utf8(data)
```
ihaveamac: Nitpick: This could be optimized by doing `util.utf16_to_utf8 = _util.utf16_to_utf8` instead. This is not essential though.

arm9/source/lua/gm9internalutil.c
```diff
@@ -0,0 +1,38 @@
...
+#include "gm9internalutil.h"
+#include "utf.h"
+#include "fs.h"
+#include "ui.h"
+
+static int internalutil_utf16_to_utf8(lua_State* L) {
+    CheckLuaArgCount(L, 1, "_util.utf16_to_utf8");
+    size_t strInLength = 0;
+    const u16* strIn = (const u16*)(void*)luaL_checklstring(L, 1, &strInLength);
+    u8 buffer[2*strInLength]; //assume a maximum of 4 utf8 codepoints per utf16 codepoint (2 bytes per utf16 codepoint so * 2)
```
ihaveamac: Wouldn't it be better to use malloc instead of this? I don't really know the constraints, but I think this could end up using too much memory for the stack.

Gruetzig: I suppose malloc would be better, but I also recently learned about luaL_buffer and I think its best to let lua handle the allocation, so I changed it to that.

data/luapackages/util.lua
Gruetzig: done

PR #973, opened 2026-07-12
Add detection for IS-RAY-DEBUGGER

### LINKED ISSUE #276: IS-CLOSER-BOX not correctly identified
Based solely off the code and the info provided on 3dbrew, IS-CLOSER-BOX units are not identified as the serial number format is unknown. This issue is simply to track that in hopes that we eventually get the serial of one.

fox8091:
This adds detection for IS-RAY-DEBUGGER units, as well as removing the erroneous note about IS-CLOSER-BOX units. Closes #276


d0k3:
Thank you!

PR #976, opened 2026-08-03
Fix PXI I2C size decode

rosaage:
During some testing I discovered changing SHMEM_BUFFER_SIZE from `2048` to `20480` broke I2C transfers.
This was due to size not being calculated correctly.

<details>
  <summary>Test code:</summary>
  
This was tested using onlinegdb.com:
  
  ```c
#include <stdio.h>
#include <stdint.h>

#define SHMEM_BUFFER_SIZE 20480
#define BIT(x) (1 << (x))

int main()
{
    uint32_t var = 0x01FF;
    var = var | (22 << 16); // "size" sent from arm9
    var = var | BIT(31); // Write operation
    uint32_t old_size = (var >> 16) % SHMEM_BUFFER_SIZE;
    uint32_t new_size = ((var >> 16) & 0x7FFF) % SHMEM_BUFFER_SIZE;

    printf("Old size: %01d, New size: %01d", old_size, new_size);

    return 0;
}
  ```
  
</details>

Output of the test code with `size` set to `22`:
`Old size: 12310, New size: 22`

The second commit adds a 15-bit size check to the arm9 side, rejecting commands early if size is too big.

### DIFF
--- a/arm9/source/system/i2c.c
+++ b/arm9/source/system/i2c.c
@@ -8,11 +8,14 @@
 bool I2C_readRegBuf(I2cDevice devId, u8 regAddr, u8 *out, u32 size)
 {
 	int ret;
-	u8 *const dataBuffer = ARM_GetSHMEM()->dataBuffer.b;
-	const u32 arg = devId | (regAddr << 8) | (size << 16);
 
 	if (size >= SHMEM_BUFFER_SIZE)
 		return false;
+	if (size > 0x7FFF)
+		return false;
+
+	u8 *const dataBuffer = ARM_GetSHMEM()->dataBuffer.b;
+	const u32 arg = devId | (regAddr << 8) | (size << 16);
 
 	ret = PXI_DoCMD(PXICMD_I2C_OP, &arg, 1);
 	ARM_InvDC_Range(dataBuffer, size);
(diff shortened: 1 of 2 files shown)


arm9/source/system/i2c.c
```diff
@@ -8,11 +8,14 @@
 bool I2C_readRegBuf(I2cDevice devId, u8 regAddr, u8 *out, u32 size)
 {
 	int ret;
-	u8 *const dataBuffer = ARM_GetSHMEM()->dataBuffer.b;
-	const u32 arg = devId | (regAddr << 8) | (size << 16);
 
 	if (size >= SHMEM_BUFFER_SIZE)
 		return false;
+	if (size > 0x7FFF)
+		return false;
```
Wolfvak: I'd rather see something like `_Static_assert(SHMEM_BUFFER_SIZE < BIT(15));` rather than the two runtime checks here, but other than that it looks perfect

the "proper" way to fix it would be to define some types in shmem.h that describe the bitfield layout but that might be a bit too many changes, I'd rather have this as it's only used once or twice

rosaage: Wouldn't this potentially not work correctly?
if SHMEM_BUFFER_SIZE is 2048 (current value) and you try to read/write 4096 bytes, then nothing will stop that with a static assert of `SHMEM_BUFFER_SIZE < BIT(31)`

